PR #26, opened 2020-07-09
add-missing-translation-labels

abdosaeedelhassan:
adding next labels:

- shownLabel
- resultsLable
- outOfLabel
- toLabel

to be used in translation for different languages

- adding whether search work: debounce or lazy

### DIFF
--- a/resources/views/includes/_options.blade.php
+++ b/resources/views/includes/_options.blade.php
@@ -21,7 +21,8 @@
         @if ($searchEnabled)
             <div class="col">
                 <input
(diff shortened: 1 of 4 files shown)


rappasoft:
The translations are redundant but I plucked the search type.

abdosaeedelhassan:
but how to translate keywords in pagination statistics?

rappasoft:
In the current version it's still a translation string you just need to use the json format. In the next version it's `laravel-livewire-tables::strings.results`

PR #36, opened 2020-08-04
Add pagination reset for perPage updates

KenLewis:
If you have a small amount of records shown, the number of pages will be bigger than when you change the perPage to a larger amount of records. For instance, if you have 50 records total with a perPage of 10 and you are on page 5, when you change the perPage to 25, there won't be any results shown. With this change, the 1st page will be shown instead of incorrectly showing the no results message.

### DIFF
--- a/src/Traits/Pagination.php
+++ b/src/Traits/Pagination.php
@@ -55,4 +55,13 @@ trait Pagination
     {
         $this->resetPage();
     }
+
+    /**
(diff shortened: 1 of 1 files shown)


amansaini419:
Is this problem resolved?

lrljoe:
> Is this problem resolved?

Yes, this was fixed back in 2020.

PR #59, opened 2020-09-10
Update composer.json

federicoromani:
Las actualizaciones en livewire 2.x no afectan éste plugin


rappasoft:
There was already one open.

PR #116, opened 2021-02-22
Add ability to filter table data

detygon:
This feature adds the ability to filter the table results, this is a work in progress. I still need to write the documentation.


rappasoft:
Just released a new version which is a ground up rebuild, column search is on my radar but this current PR will no longer work. Appreciate your hard work though!

PR #199, opened 2021-04-15
Feature 1.0.0

bdelamatre:
- Added row-columns blade and set rowView() method to this new blade. This makes the definition of a rowView blade optional per-table
- re-added format() method on column and with new parameter ordering. new parameter ordering is meant to make it easier to grab only the needed information on a callback (value, column, row)
- re-added formatted() method to column with some enhancements to make it easier to use. This is so that formatted() can be called indiscriminately from a blade template
- added asHtml() mutator to control data output. this was also an issue in the primary branch.
- re-added support for getTableRowUrl link
- flipped column and text parameters for Column constructor to improve backwards compatibility, if column parameter is missing then set it to the column text snake cased.


rappasoft:
Made a couple tweaks but looks great! I think it's good to have that default functionality. 

Thanks again!

PR #210, opened 2021-04-18
Adding searchable() and out-of-box search

bdelamatre:
The following patches are included:
* pagination is now optional and may be disabled using the $pagination property
* added searchable() method back to columns
* added applySearchFilter() method which provides out-of-box search functionality for columns indicated as searchable()
* added type(), boolean(), number() and text() mutators for columns to provide additional specification. will slightly alter search functionality.
* added aggregate() mutator for columns to provide additional specification. will skip search functionality.
* added ColumnUtilities class which adds some functions for inspecting columns on a builder
* will attempt to map eloquent relations to previously selected fields (for cases where manual selections/joins are made)
* added first typed exception

minimally tested in local project


rappasoft:
Oh boy, this is a lot of stuff. My original intention with this version was to get away from the column builder so people had more control, but I'll see if I can wrap my head around what you have here. Thanks!

bdelamatre:
I like that you can interact directly with the search filter from the query and I think that's useful for complex searching. On the other hand, same with the rowView, I think the package should make some effort to search with minimal configuration. 

I actually quite like the Column builder, it's familiar coming from projects like Nova. We have enough information from the column definition to handle 80% of scenarios and allow people to start producing tables quickly.

rappasoft:
I'll be honest there's so much in here I don't even know where to start. Might have to try to break it into more manageable pieces somehow.

bdelamatre:
1. New mutators on Columns

searchable() - works pretty much just like before
type() - specify a column's type (text, number and boolean supported). default is text.
number() - helper to set type() of number
boolean() - helper to set type() of boolean
aggregate() - specify column as an aggregate (e.g. count, sum, min, max, ...)

2. New utility class of ColumnUtilities which offers functions to inspect a Query Builder instance for certain pieces of information. This is somewhat similar in utility to the Yajra.php trait that was dropped. The main purpose of this utility is to try and map a column defined on the table to a column from the select statement. 

3. applySearchFilter() which works very similar to the previous search functionality. It loops through the column definition, picks out columns that are searchable() and not aggregate(). 

It then checks if those columns were selected on the query (using ColumnUtilities). 

a. If the column was selected on the query, or not on a relation, we use orWhere() since it can reference that selected column
b. If the column wasn't selected on the query, we lean on orWhereHas() from the Eloquent Query Builder to map a column on a relation.

We then use slightly different matching logic based on the column type.

text = %.$search.%
number = $search
boolean = (bool)$search

All the other little things, like the pagination checks, etc. are noise.

bdelamatre:
I broke out pagination on https://github.com/rappasoft/laravel-livewire-tables/pull/222

I will leave this pull request open for now, but I intend to submit a replacement pull request(s) targeting just search.

rappasoft:
Yeah, that's fine. More manageable chunks are good.  I will probably get into a weekly release cycle once we settle down and have a solid feature set. I don't mind doing a lot now since it's all just value add to the package, which you are a huge help with and I thank you.

bdelamatre:
https://github.com/rappasoft/laravel-livewire-tables/pull/233

PR #211, opened 2021-04-18
Search model options

bdelamatre:
Adds properties to compute model options on search property. This is presently set to .debounce.250ms in themes. This allows user to control those options without overriding the themes.

### DIFF
--- a/resources/views/bootstrap-4/includes/search.blade.php
+++ b/resources/views/bootstrap-4/includes/search.blade.php
@@ -1,7 +1,7 @@
 @if ($showSearch)
     <div class="mb-3 mb-md-0 input-group">
         <input
-            wire:model.debounce.250ms="filters.search"
+            wire:model{{ $this->searchFilterOptions }}="filters.search"
             placeholder="{{ __('Search') }}"
             type="text"
             class="form-control"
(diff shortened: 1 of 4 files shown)


rappasoft:
Should we be able to include the default of debounce 250ms instead of removing it?

bdelamatre:
I don't have an opinion on what the package default should be. If undefined, the livewire default is 150ms.

My own testing showed that:
* 250ms wasn't enough of a debounce to prevent misfires while typing
* On slower queries, misfires produced undesirable results.
* Lazy loading was the most reliable method for slower queries as it eliminated misfires. However, it's not entirely clear from a UI perspective that you need move off the input to fire a search.

rappasoft:
What if we just added 1 property of `?string $searchFilterProperty = null` or something that the users would just add `.lazy` or `.debounce.250ms` or whatever they wanted? It would basically just append to the wire:model, if Livewire added different properties we wouldn't have to add anything. Am I missing a side effect of doing it this way?

bdelamatre:
That would work. Pros would be as you mentioned and it eliminates possible ambiguity in how options are preferred. Cons are that the options may be less evident and an increased risk that an incorrect value could break livewire.

I would ltrim the dot and manually add it to reduce that risk.

Do you want me to submit a different pull?

rappasoft:
That's true both ways, I want to keep the codebase as small as possible for maintainability. They are both good options, the only confusion I would see with the 3 properties is someone accidentally using more than 1, and not understand which order they are being processed. Either way, we just need to document it well.

I would say the more Laravel way would be separating them into multiple properties for readability. So I think it's good, I'll pull and update the readme tonight when I get home.

PR #212, opened 2021-04-18
fix cleanFilter() bug and auto-initialize filters parameter

bdelamatre:
* commented out $this->cleanFilters() on getRowsQueryProperty(). Filtering was broken in my environment and this was the cause. I couldn't figure out why this would be here so I commented it out.
* added mountWithFilters which will auto-initialize $filters from filters(). This removes the requirement to define $filters and filters().

### DIFF
--- a/resources/views/tailwind/includes/filters.blade.php
+++ b/resources/views/tailwind/includes/filters.blade.php
@@ -51,6 +51,7 @@
                                 <div class="mt-1 relative rounded-md shadow-sm">
                                     <select
                                         wire:model="filters.{{ $key }}"
+                                        wire:key="filter-{{ $key }}"
                                         id="filter-{{ $key }}"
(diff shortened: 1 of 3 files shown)


rappasoft:
https://github.com/rappasoft/laravel-livewire-tables/blob/master/src/Traits/WithFilters.php#L40 cleanFilters makes sure the option being selected exists in the list of options you allow, to prevent any SQL security issues.

bdelamatre:
Ahh, that gives me more to go off, thanks. I bet there is something like a typecasting mismatch. I will dig further.

You can still define $filters if you have defaults, it only defines them on mount if they are missing. I suppose we could add a default field to the fields() definition. It's not necessary, but that would keep the entire filters definition in one place.

bdelamatre:
Replaced with https://github.com/rappasoft/laravel-livewire-tables/pull/213

PR #213, opened 2021-04-19
Refactor cleanFilter() and auto-initialize filters parameter

bdelamatre:
* Refactored cleanFilter() to filter out all keys that don't have a filters() definition
* When cleaning filters, force typecasting on integer values
* added mountWithFilters which will auto-initialize $filters from filters(). This removes the requirement to define $filters and filters().

I wasn't able to find a proper way in livewire to typecast these values, hence the typecasting when a specific type is checked. I am sure there is a better way to handle this, but this works.

### DIFF
--- a/resources/views/tailwind/includes/filters.blade.php
+++ b/resources/views/tailwind/includes/filters.blade.php
@@ -51,6 +51,7 @@
                                 <div class="mt-1 relative rounded-md shadow-sm">
                                     <select
                                         wire:model="filters.{{ $key }}"
+                                        wire:key="filter-{{ $key }}"
                                         id="filter-{{ $key }}"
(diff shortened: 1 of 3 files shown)


rappasoft:
Made some slight tweaks to cleanFilters but I don't think I changed any of what you were after.

rappasoft:
https://user-images.githubusercontent.com/6026504/115173427-4568cf00-a095-11eb-8ad3-8d6aa03a0457.mov

Also currently have an issue with multiple filters and removing the pills. Seems to be a conflict with the wire:key's. I renamed them all to be unique but I'm still having issues.

rappasoft:
Think I fixed it with this commit: https://github.com/rappasoft/laravel-livewire-tables/commit/392e7d5a7162b96be0216e15ef4ce5b0720b4816

rappasoft:
I don't know why this didn't get merged with the last dev->master but the code is in there.

PR #216, opened 2021-04-19
tap into the sort query

devsrv:
this will allow users to customize the sort query like below

> in user's livewire table component
```php
public function tapSorting(Builder $query, $field, $direction) {
    if($field === 'company_name') {
        $sortQuery = Company::select('name')
                    ->whereColumn('companies.user_id', 'users.id'); // a custom query for orderBy(..)

        return $query->orderBy($sortQuery, $direction);
    }
}
```

### DIFF
--- a/src/Traits/WithSorting.php
+++ b/src/Traits/WithSorting.php
@@ -30,9 +30,14 @@ trait WithSorting
     public function applySorting(Builder $query): Builder
     {
         foreach ($this->sorts as $field => $direction) {
+            if(! $this->tapSorting($query, $field, $direction))
             $query->orderBy($field, $direction);
         }
 
+        return $query;
+    }
+    
+    public function tapSorting(Builder $query, $field, $direction) {
         return $query;
     }
 


rappasoft:
I like the idea, however there might be a better way to do it.

If we are going to go this route we can probably add a sorting() callback to the Column itself, much like the old version.

Let me think about it a little bit.

devsrv:
callback will be great (y)

bdelamatre:
Am I the only one that thinks tap implies registering a callback (or multiple)? E.g. if we have a tapSorting() I would expect it to register a callback that accepts (and returns) the builder during sorting. We then would be able to define multiple taps either on a property or by injecting them at some point in the class lifecycle (construct?, mount?, somewhere else?)

If we get this right here we could probably apply the same pattern to search/filter/etc.

rappasoft:
Maybe something like: (untested)

```
public function sortable(callable $callback = null): self
{
    $this->sortable = true;

    if ($callback) {
        $this->sortCallback = $callback;
    }

    return $this;
}
```

```
public function applySorting(Builder $query): Builder
{
    foreach ($this->sorts as $field => $direction) {
        if ($this->getColumn($field)->hasSortCallback()) {
            $query = app()->call($this->getColumn($field)->getSortCallback(), ['builder' => $query, 'direction' => $direction]);
        } else {
            $query->orderBy($field, $direction);
        }
    }

    return $query;
}
```

Then the above example:

```
Column::make(__('Name'))
    ->sortable(function(Builder $builder, string $direction) {
        return $builder->orderBy(Company::select('name')->whereColumn('companies.user_id', 'users.id'), $direction);
    }),
```

devsrv:
> ```
> Column::make(__('Name'))
>     ->sortable(function(Builder $builder, string $direction) {
>         return $builder->orderBy(Company::select('name')->whereColumn('companies.user_id', 'users.id'), $direction);
>     }),
> ```

its neat 👌

rappasoft:
So I have it working on this branch with your example: https://github.com/rappasoft/laravel-livewire-tables/commit/cffd2ef0026b5c35269e172896cbd42701ed30b7

rappasoft:
WIll merge this new feature into the next release!

PR #218, opened 2021-04-20
testing

bdelamatre:
After thinking through different ways to approach unit testing for the project, this is my attempt to add an initial testing suite. The test workflow is:

* TestCase->setUp() boots up and creates a simple test database using sqlite
* models and a datatable are created to match the test database
* interactions are tested directly against the PetsTable component

The approach is simple in that it:
* can be ran from within the project with no external dependencies
* doesn't require bootstrapping an entire demo project

This approach is currently limited in that it:
* can't execute Livewire / HTTP / Browser tests

An example test looks like this:
```   
/** @test */
public function test_search_filter()
{
        $this->table->filters['search'] = 'Cartman';
        $this->assertEquals(1, $this->table->getRowsProperty()->total());
}
```


rappasoft:
This is great thanks! I made some changes which you can see here: https://github.com/rappasoft/laravel-livewire-tables/commit/bff37147ca8887336247752ccd155f52b38da0bf

I think it will be a good start, I appreciate all the effort you've been putting in to help with this package.

PR #222, opened 2021-04-22
Feature optional pagination

bdelamatre:
Allows pagination to be controlled using the $pagination property. True by default. Setting $pagination = false will disable pagination throughout.

Added tests.


rappasoft:
Changed to $paginationEnabled in case we have more things to turn on/off we can use the convention. Thanks for writing tests!

rappasoft:
All of a sudden psalm doesn't like the magic properties: https://github.com/rappasoft/laravel-livewire-tables/runs/2410528328?check_suite_focus=true#step:6:8

I don't know how to get it to stop worrying about that now. Weird that it didn't before.

rappasoft:
This fixes it and should be backward compatible? We should probably steer away from the magic properties as much as we can. https://github.com/rappasoft/laravel-livewire-tables/commit/e0da5160de8bd5df70fe5ef26ad7aa992d259e6e

PR #233, opened 2021-04-24
new feature: searchable()

bdelamatre:
Follow-up to https://github.com/rappasoft/laravel-livewire-tables/pull/210

* re-added searchable() method to columns
* added applySearchFilter() which applies default search behavior to searchable() columns
* utilities class to replace pre 1.0 trait functionality
* tests

### DIFF
--- a/src/DataTableComponent.php
+++ b/src/DataTableComponent.php
@@ -123,15 +123,19 @@ abstract class DataTableComponent extends Component
     {
         $this->cleanFilters();
 
-        return $this->applySorting($this->query());
-    }
+        $query = $this->query();
 
-    /**
-     * @return Builder
-     */
-    public function getRowsQueryProperty(): Builder
-    {
-        return $this->rowsQuery();
+        // sorting?
+        if (method_exists($this, 'applySorting')) {
+            $query = $this->applySorting($query);
+        }
+
+        // searching?
+        if (method_exists($this, 'applySearchFilter')) {
+            $query = $this->applySearchFilter($query);
+        }
+
+        return $query;
     }
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to `laravel-livewire-tables` will be documented in this file
 
 ## [Unreleased]
 
+## [1.2.0] - 2021-04-22
+
 ### Added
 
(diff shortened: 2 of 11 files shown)


rappasoft:
Gonna need to have some actual documentation after all these features we're plowing into this thing.

src/DataTableComponent.php
```diff
@@ -123,15 +123,19 @@ public function rowsQuery(): Builder
...
+        $query = $this->query();
 
-    /**
-     * @return Builder
-     */
-    public function getRowsQueryProperty(): Builder
-    {
-        return $this->rowsQuery();
+        // sorting?
+        if (method_exists($this, 'applySorting')) {
```
rappasoft, line 129: Shouldn't these two methods always exist since they are part of the package and not user defined?

bdelamatre: They are defined in traits, so as long as the traits are applied to the class. I could go either way given this implementation if you have a preference.

In general I assume that traits are meant to be mixed and matched. So, maybe there is a future scenario where someone would need a table component but without specific traits. 

E.g. I could see on https://github.com/rappasoft/laravel-livewire-tables/discussions/200 the possibility of creating a separate component for collections which may call for excluding these traits (or mixing in others).

rappasoft: True, though in the case of this package even though the traits are functional, they were more meant for organizational purposes. Having them doesn't hurt though.

rappasoft:
Ok I have some cleanup done locally and everything is working well, I actually can't find a need for the old way with my test table. The only thing I can't get to work is the searchable closure, no matter what I add in there the search just returns nothing altogether. Since you have a better idea of how it works can you just see if it's working correctly, and then maybe put a little blurb in the readme?

rappasoft:
Since I merged this into develop, if you end up finding an issue or want to add to the readme I'll just merge a new PR if you don't mind. I'm going to try to dig into why I can't get it to work tomorrow.

This was a great PR by the way that's a lot of work and we all appreciate it.

bdelamatre:
Thanks!

I see the issue with callback and submitted a separate PR:
https://github.com/rappasoft/laravel-livewire-tables/pull/234

However, I missed your note there about updating the readme.

PR #238, opened 2021-04-25
filter label

bdelamatre:
Modified views to grab the filter labels for filter values.

```
@if(isset($customFilters[$key]) && method_exists($customFilters[$key], 'options'))
    {{ $customFilters[$key]->options()[$value] ?? $value }}
@else
    {{ ucwords(strtr($value, ['_' => ' ', '-' => ' '])) }}
@endif
```

I imagine this should be the expected default behavior, but we could add an option if not.

### DIFF
--- a/resources/views/bootstrap-4/includes/filter-pills.blade.php
+++ b/resources/views/bootstrap-4/includes/filter-pills.blade.php
@@ -8,7 +8,12 @@
                     wire:key="filter-pill-{{ $key }}"
                     class="badge badge-pill badge-info d-inline-flex align-items-center"
                 >
-                    {{ $filterNames[$key] ?? ucwords(strtr($key, ['_' => ' ', '-' => ' '])) }}: {{ ucwords(strtr($value, ['_' => ' ', '-' => ' '])) }}
+                    {{ $filterNames[$key] ?? ucwords(strtr($key, ['_' => ' ', '-' => ' '])) }}:
+                    @if(isset($customFilters[$key]) && method_exists($customFilters[$key], 'options'))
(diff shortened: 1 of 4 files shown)


rappasoft:
I'm confused about the use case lol. Don't we just use the values of the array to display anyway? I'm not against it I just want to understand what your thought process is.

bdelamatre:
The option value is what's stored in the filter and presently displayed. For many of my use cases that value will be different from the option content/label in the dropdown.

For example, I will have a dropdown generated from pairs of keys (option value) and display values (option label). The key is the value and what's used by the query for filtering. The display value is what the user sees.

1 = Pending
2 = Error
3 = Complete

Or it could even be a simple boolean option

0 = No
1 = Yes

rappasoft:
Can you give me example code on how to use it for the readme?

bdelamatre:
Consider this example from the docs

```
 'type' => Filter::make('User Type')
            ->select([
                '' => 'Any',
                User::TYPE_ADMIN => 'Administrators',
                User::TYPE_USER => 'Users',
            ]),
```

User::TYPE_ADMIN and User::TYPE_USER are likely either a) constants like `admin` and `user` or b) database keys like 1 and 2.  The filter pill would be transformed from:

Type: admin -or- Type: 1 

To  match the select dropdown

Type: Administrators

No changes are required to implement this, it just works and fails back to displaying `admin` or `1` if it can't find the select option for some reason.

rappasoft:
For some reason, I'm having a brain fart on how to implement $customFilters, is it an array or being used as a method call in the livewire dom? My confusion comes from the use of the option method call on the array.

PR #239, opened 2021-04-25
fix results count on tailwind theme

bdelamatre:
Ensures that a results count is displayed even when a) there is less than one page of results and b) pagination is disabled

### DIFF
--- a/resources/views/tailwind/includes/pagination.blade.php
+++ b/resources/views/tailwind/includes/pagination.blade.php
@@ -1,5 +1,7 @@
-@if ($paginationEnabled && $showPerPage)
-    <div class="p-6 md:p-0">
+<div class="p-6 md:p-0">
+    @if ($paginationEnabled && $showPerPage && $rows->lastPage() > 1)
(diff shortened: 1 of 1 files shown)


rappasoft:
Hmm we should probably do this for all of the themes.

bdelamatre:
A quick smoke test and looked like this was already working somehow for bootstrap and the current pagination views. Not sure why the difference.

rappasoft:
Hmm for me pagination does not show lol. I'll dig into it.

rappasoft:
Bootstrap 4 and 5 show nothing for me if paginationEnabled is false, also just noticed, tailwind pagination has the results and links reversed from that Bootstrap has.

molnarjg:
@rappasoft Is it possible to only show either the pager buttons OR the "Showing .." text with the pagination buttons on the right? Having them all at once with Tailwind is a little bit too much:
image

rappasoft:
It's normally just the right side of your screenshot. I don't know how you got both simple and advanced pagination to show up at once?

PR #267, opened 2021-05-03
Set "selectAll", if the current count is all

spekulatius:
Hello @rappasoft,

I've noticed the bulk message being incorrect if there is only one page of entries:

image

This PR should fix this.

Cheers,
Peter

### DIFF
--- a/src/Traits/WithBulkActions.php
+++ b/src/Traits/WithBulkActions.php
@@ -44,6 +44,8 @@ trait WithBulkActions
     public function selectPageRows(): void
     {
         $this->selected = $this->rows->pluck($this->primaryKey)->map(fn ($key) => (string) $key);
+
+        $this->selectAll = count($this->selected) === $this->rows->count();
     }
 
     public function selectAll(): void


rappasoft:
This actually won't work right, If there are say 5 pages, and the last page has 1 entry and you select the top checkbox, it will select all entries on all pages. This only works if there is only 1 page with less than perPage entries.

I actually changed this sometime yesterday on develop in a different way: https://github.com/rappasoft/laravel-livewire-tables/commit/ebe5de929bbeacb7a0ed11e9b072556035643883

spekulatius:
Oh, I might have put my arrow in the wrong spot (and explained it insufficient). I didn't want to refer to the "select all" - more to the "1 of 1" selected and the "select all" being still an option (while it shouldn't be).

rappasoft:
Right i'm saying the implementation has side effects as far as I've tested. The commit I linked was my solution to this exact problem on dev yesterday or so. If you make a table with a few pages and have the last page have less than the page limit (i.e. seed 15 records at 10 per page) and you select the page checkbox on the last page, it will also select the first page, which is wrong.

The way I did it I didn't change the backend, just modified the UI to be more clear.

spekulatius:
Okay, that makes sense. Let's close this :+1:

PR #279, opened 2021-05-10
Allows to use Relation instead of Builder to generate data

fabio-ivona:
During my development using this awesome package, I needed to display a table taking data from a many to many relationship:

Course <=> Subscription <=> Student

in order to render a table to display all students of a specific course, I used laravel-livewire-tables and tried to take data this way

```php
class StudentstTable extends DataTableComponent
{
   //.....

   public function query(): Builder
    {
        return $this->course->students()->getQuery();
    }

  //.....
}
```

I thought everything worked until I tried to take one of the row's ID and found that it was not the ID of the Student, but the ID of the pivot table (Subscription)

that's because getQuery() of a many to many relationship mixes columns of both tables

an elegant solution would be to return an `Illuminate\Database\Eloquent\Relations\Relation` instead of the Builder:

```php
class StudentstTable extends DataTableComponent
{
   //.....

   public function query(): Builder|Relation
    {
        return $this->course->students();
    }

  //.....
}
```

but this is not possible because of the strict return type checking of the query() method's signature

this PR slightly changes the code in order to loose the strict type checking, allowing to return a Relation. 

every test runs smoothly without any change and this is a non breaking change: removing typechecking from parent method allows to still use typechecking on the classes extending it.

I could not use union types because of the php7.4 requirement, maybe this would be a nice additional PR in a future version as the php7.4 requirement will be dropped

@rappasoft sorry for my english, I'm not a native speaker. I hope to have well explained my goal with this pull request

and thanks for you awesome work! You saved me a lot of code!

edit: **added tests**

### DIFF
--- a/src/DataTableComponent.php
+++ b/src/DataTableComponent.php
@@ -2,6 +2,7 @@
 
 namespace Rappasoft\LaravelLivewireTables;
 
+use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Support\Collection;
@@ -99,9 +100,9 @@ abstract class DataTableComponent extends Component
     /**
      * The base query with search and filters for the table.
      *
-     * @return Builder
+     * @return Builder|Relation
      */
-    abstract public function query(): Builder;
+    abstract public function query();
 
     /**
      * TableComponent constructor.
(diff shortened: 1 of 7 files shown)


rappasoft:
I don't see a problem with this. The test suite is not that powerful, so I'll try to put it through some manual tests to make sure nothing breaks.

fabio-ivona:
@rappasoft may I help by adding some test for a specific aspect that worries you?

rappasoft:
No this is fine I just have nothing else to release with it right now. I'll add a patch tomorrow if nothing else comes through.

fabio-ivona:
great!

again! thanks for your awesome and valuable work!

PR #282, opened 2021-05-14
Adjusts Table Header Items Alignment to the start

viniqueiroz:


rappasoft:
This is supposed to be center. Thanks though.

PR #283, opened 2021-05-14
Fix ambiguous column id when using Relation instead of Builder

fabio-ivona:
@rappasoft I noticed a slight issue when using bulk actions with a many to many Relation to generate data:

if the pivot table has an **id** field, the query will generate an _integrity constraint violation_ when using the **id** column to filter data:

```
SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in field list is ambiguous (SQL: select `id` from `students` inner join `subscriptions` on `students`.`id` = `subscriptions`.`student_id` where `subscriptions`.`edition_id` = 3 and `students`.`id` in (3) order by `students`.`last_name` asc)
```

this happens when executing the `WithBulkActions::selectedRowsQuery()` and `WithBulkActions::selectedKeys()` methods and can be avoided by letting laravel to qualify the _primaryKey_ column:

`$this->primaryKey //id` 

would become:

`$query->qualifyColumn($this->primaryKey) //students.id`  

this is a very rare case, but could happen

### DIFF
--- a/src/Traits/WithBulkActions.php
+++ b/src/Traits/WithBulkActions.php
@@ -65,7 +65,7 @@ trait WithBulkActions
     public function selectedRowsQuery()
     {
         return (clone $this->rowsQuery())
-            ->unless($this->selectAll, fn ($query) => $query->whereIn($this->primaryKey, $this->selected));
+            ->unless($this->selectAll, fn ($query) => $query->whereIn($query->qualifyColumn($this->primaryKey), $this->selected));
     }
 
     /**
@@ -78,7 +78,7 @@ trait WithBulkActions
 
     public function selectedKeys(): array
     {
-        return $this->selectedRowsQuery()->pluck($this->primaryKey)->toArray();
+        return $this->selectedRowsQuery()->pluck($this->rowsQuery()->qualifyColumn($this->primaryKey))->toArray();
     }
 
     public function getSelectedKeysProperty(): array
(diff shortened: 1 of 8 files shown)


fabio-ivona:
@rappasoft seems that automatic tests are cancealed by some github policy

src/Traits/WithBulkActions.php
```diff
@@ -65,7 +65,7 @@ public function resetBulk(): void
     public function selectedRowsQuery()
     {
         return (clone $this->rowsQuery())
-            ->unless($this->selectAll, fn ($query) => $query->whereIn($query->qualifyColumn($this->primaryKey), $this->selected));
```
rappasoft: Was there a reason you removed this one? Also, is there qualifyColumn documentation?

fabio-ivona: damn, I forgot to undo it, it was a try because I noticed that the test I wrote was succeeding with the old code. so I put the PR in draft while was writing a better test

just fixing it

fabio-ivona: anyway, qualifyColumn is not documented in official Laravel documentation, it was released with laravel 5.5.29:

https://laravel-news.com/laravel-5-5-29-released

but it simply does this:

```
public function qualifyColumn($column)
{
  return $this->model->qualifyColumn($column);
}
```

and Model::qualifyColumn() adds the table name before the column name:

```
public function qualifyColumn($column)
{
    if (Str::contains($column, '.')) {
        return $column;
    }

    return $this->getTable().'.'.$column;
}
```

fabio-ivona:
@rappasoft I just rewrote my test table and RelationshipDatatableComponentTest in order to have an example that gives an error without this PR and it is fixed by adding qualifyColumn()

PR #284, opened 2021-05-14
fix tailwind pagination view

bdelamatre:


rappasoft:
What's the use case here? Also, the other templates would be unchanged but have the same issue as you are trying to fix. But I feel like we've gone back and forth with the wrapping if statement and I don't remember why haha.

bdelamatre:
The use case is that I have a table, with pagination, but I don't need to show the per-page dropdown. The current behavior is If showPerPage is false then pagination links are hidden.

I forget as well. I think the last time I touched it the goal was to display the result count regardless of whether pagination was enabled or not. This behavior was a little different on the tailwind theme since the bootstrap links() displayed the count reliably while tailwind didn't.

rappasoft:
What about the [$showPerPage](https://github.com/rappasoft/laravel-livewire-tables/wiki/Display) property?

bdelamatre:
I would expect the showPerPage property to be limited to showing/hiding the
per page drop-down.

rappasoft:
What if we just changed it to:

```html
@if ($showPagination)
    <div class="p-6 md:p-0">
        @if ($paginationEnabled && $rows->lastPage() > 1)
            {{ $rows->links() }}
        @else
            <p class="text-sm text-gray-700 leading-5">
                @lang('Showing')
                <span class="font-medium">{{ $rows->count() }}</span>
                @lang('results')
            </p>
        @endif
    </div>
@endif
```

bdelamatre:
Sure.

PR #285, opened 2021-05-14
hide filters when $showFilters is false

bdelamatre:
Pretty much the title. 

Scenario is that you have a table with filters (e.g. controlled via query parameters or another form) but you don't want the filters to show. I would expect $showFilters to hide the filters dropdown, but it currently doesn't.

### DIFF
--- a/resources/views/tailwind/includes/filters.blade.php
+++ b/resources/views/tailwind/includes/filters.blade.php
@@ -1,4 +1,4 @@
-@if ($filtersView || count($customFilters))
+@if ($showFilters && ($filtersView || count($customFilters)))
     <div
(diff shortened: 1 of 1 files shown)


rappasoft:
Hmm, tricky because $showSorting hides the pills but does not disable sorting, not sure if it's completely the same but I want to make sure I don't confuse people.

bdelamatre:
Could call for a new showFilterForm property if you think.

rappasoft:
Yeah, I think a different property, only because I use the default dropdown, but I do hide the pills with that sometimes as they're mostly fluff. `$showFilterDropdown` or something.

PR #286, opened 2021-05-17
Use column text for sorting and filter pills if no $filterNames or $sortNames exist

fabio-ivona:
This PR aims to define a better default value for filter/sorting pills, by giving precedence on the column text (if exists) over the column name of the model

let's see a use case: in non-english languages db column names often differs from what's displayed to the user, and this is greatly handled by laravel-livewire-tables:

```
public function columns(): array
{
    return [
        Column::make('Email', 'email')->sortable()->searchable(),
        Column::make('Nome', 'first_name')->sortable()->searchable(),
        Column::make('Cognome', 'last_name')->sortable()->searchable(),
        Column::make('', 'actions'),
    ];
}
```

but, for filtering and sorting I obtain this:

image

that's to say the human readable db column name: `ucwords(strtr('first_name', ['_' => ' ', '-' => ' '])) // First Name`

but, usually, I want the filter pills labels to be equal to the displayed column name:

image

I know this could solved by defining `$filterNames` and `$sortNames` properties, but that's kinda redundant.. as my labels are already written in columns definitions

so I think we can create a "pills label" priority such as

`$sortNames[$col]` > Column Text > Human Readable Db Column 

this can be obtained by changing this:

```
{{ $sortNames[$col] ?? ucwords(strtr($col, ['_' => ' ', '-' => ' '])) }}
```

to this:

```
{{ $sortNames[$col] ?? collect($this->columns())->pluck('text', 'column')->get($col, ucwords(strtr($col, ['_' => ' ', '-' => ' ']))) }}
```

hope to have well explained the issue and the solution I'm trying to propose

_a better solution would be extracting this logic in a method of `WithFilter` and `WithSorting` traits, but did not want to change so many lines without your opinion about this_

as usual, thumbs up for your great work!

### DIFF
--- a/resources/views/bootstrap-4/includes/filter-pills.blade.php
+++ b/resources/views/bootstrap-4/includes/filter-pills.blade.php
@@ -8,7 +8,7 @@
                     wire:key="filter-pill-{{ $key }}"
                     class="badge badge-pill badge-info d-inline-flex align-items-center"
                 >
-                    {{ $filterNames[$key] ?? ucwords(strtr($key, ['_' => ' ', '-' => ' '])) }}:
+                    {{ $filterNames[$key] ?? collect($this->columns())->pluck('text', 'column')->get($key, ucwords(strtr($key, ['_' => ' ', '-' => ' ']))) }}:
                     @if(isset($customFilters[$key]) && method_exists($customFilters[$key], 'options'))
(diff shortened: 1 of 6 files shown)


rappasoft:
Very good thanks!

PR #301, opened 2021-05-31
Enhance search filter

lloricode:
Hi @rappasoft 

If this good idea to able to search multiple result in between space

thank you

### DIFF
--- a/src/Traits/WithFilters.php
+++ b/src/Traits/WithFilters.php
@@ -287,6 +287,12 @@ trait WithFilters
                     // Let's try to map this column to a selected column
                     $selectedColumn = ColumnUtilities::mapToSelected($column->column(), $query);
 
+                    $searchEach = function(Builder $query, string $cloumn, string $search){
+                        foreach (explode(' ', $search) as $value) {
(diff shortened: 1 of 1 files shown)


lloricode:
if this ok, i will try to fix test

rappasoft:
I think you can already do this using custom search scopes without the need to update the core. Just explode the term by the space and loop it.

lloricode:
> I think you can already do this using custom search scopes without the need to update the core. Just explode the term by the space and loop it.

Thanks for idea, but i have something, can we add configuration in parent class for this? something like separate function so we able to override it in our local base class extending by your?

rappasoft:
The problem is most people just want to search from the string they are entering, whether it be more than one word or not. I think there would be confusion if you're getting multiple search results from one search term because it's like doing `N` different searches.

lloricode:
> The problem is most people just want to search from the string they are entering, whether it be more than one word or not. I think there would be confusion if you're getting multiple search results from one search term because it's like doing `N` different searches.

I see, good point and I understand now, thanks

PR #304, opened 2021-05-31
Suggestion: adding .stop to select filters

stephengalbraith:
First off @rappasoft - great work on this package! I've been using this and learning a lot from it.

I realized that when applying a filter in the dropdown component, the dropdown would close. I would suggest generally that often times multiple filters may get applied so it doesn't make sense that it closes on each change of the select box.

I'm proposing adding the ".stop" modifier on the select boxes. I wasn't sure if the Bootstrap templates needed that extra `onclick="event.stopPropagation();"` so I removed then in this commit.

### DIFF
--- a/resources/views/bootstrap-4/includes/filters.blade.php
+++ b/resources/views/bootstrap-4/includes/filters.blade.php
@@ -24,8 +24,7 @@
                                 </label>
 
                                 <select
-                                    onclick="event.stopPropagation();"
-                                    wire:model.prevent="filters.{{ $key }}"
+                                    wire:model.stop="filters.{{ $key }}"
                                     id="filter-{{ $key }}"
                                     class="form-control"
                                 >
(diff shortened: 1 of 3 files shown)


rappasoft:
I didn't even know .stop was available, I think that's what I was trying to accomplish with the javascript. I can't even find it in the documentation. If it works I'll add it to the next release.

stephengalbraith:
Either .stop or .prevent seems to work on what I am working on. [Livewire docs for event modifiers](https://laravel-livewire.com/docs/2.x/actions#event-modifiers)

rappasoft:
So I can't get this to work like you say it does for you. For BS4, it won't let me select. For BS5, it lets  me choose but still closes:

https://user-images.githubusercontent.com/6026504/120909287-a21e4a00-c641-11eb-9bc8-6f2751dd5946.mov

stephengalbraith:
I will look into this a bit more as soon as I can for you!

rappasoft:
We fixed this in the date filter PR. Thanks though!

PR #306, opened 2021-06-01
Additng the sortable option

Greenvold:
It would be great to add sortable option to this package via config setup. I am trying to implement this tables with sortable packages and it would help a lot.


rappasoft:
What does this mean?

PR #316, opened 2021-06-05
adding exported property

abdosaeedelhassan:


rappasoft:
For?

abdosaeedelhassan:
To set the column available for exporting or not

rappasoft:
Why would this be needed? It would not be used anywhere?

abdosaeedelhassan:
we can used like next
`  Column::make(__('Actions'))
                ->format(function ($value, $column, $row) {
                   ....
                    return $html;
                })->asHtml()->exported(false)`

to determine this column is available for exporting to (pdf, excel, ...) or not

PR #372, opened 2021-06-23
Remove Colspan on bulk-select-row.blade.php

rizkhal:
I dont see property colspan on component.
I think you forgot to remove them.


rappasoft:
[It's right here on every theme](https://github.com/rappasoft/laravel-livewire-tables/blob/master/resources/views/tailwind/includes/table.blade.php#L41). It's also not causing an issue that I can see, this is also an empty pull request.

PR #377, opened 2021-06-23
add empty message to lang

rizkhal:
empty message on `DataTableComponent` in lang


rappasoft:
This is configurable per table: https://github.com/rappasoft/laravel-livewire-tables/wiki/Display

rizkhal:
I think it will take some time to add `$emptyMessage` in each component if it has multiple table components while messages on all tables will only contain the same message

PR #408, opened 2021-07-20
Add make:table command

iAmKevinMcKee:
I've added the `make:table` command. You just pass a class name and it scaffolds out a stub that helps you get started.

Hopefully this is pretty self explanatory. I used a lot of the code and approach as the Livewire make command.

Sorry I don't have any tests.

### DIFF
--- a/src/LaravelLivewireTablesServiceProvider.php
+++ b/src/LaravelLivewireTablesServiceProvider.php
@@ -3,6 +3,7 @@
 namespace Rappasoft\LaravelLivewireTables;
 
 use Illuminate\Support\Facades\Blade;
+use Rappasoft\LaravelLivewireTables\Commands\MakeCommand;
 use Spatie\LaravelPackageTools\Package;
 use Spatie\LaravelPackageTools\PackageServiceProvider;
 
@@ -32,6 +33,17 @@ class LaravelLivewireTablesServiceProvider extends PackageServiceProvider
         Blade::component('livewire-tables::bootstrap-5.components.table.heading', 'livewire-tables::bs5.table.heading');
         Blade::component('livewire-tables::bootstrap-5.components.table.row', 'livewire-tables::bs5.table.row');
         Blade::component('livewire-tables::bootstrap-5.components.table.cell', 'livewire-tables::bs5.table.cell');
+
+        $this->registerCommands();
+    }
+
+    protected function registerCommands()
+    {
+        if (! $this->app->runningInConsole()) return;
+
+        $this->commands([
+            MakeCommand::class,         // make:table
+        ]);
     }
 
     /**
--- a/src/Commands/MakeCommand.php
+++ b/src/Commands/MakeCommand.php
@@ -21,7 +21,7 @@ class MakeCommand extends Command
     protected $signature = 'make:table
         {name : The name of your Livewire class}
         {model? : The name of the model you want to use in this table }
-        {--V|view : We will generate a row view for you}
+        {--view : We will generate a row view for you}
         {--force }';
 
     /**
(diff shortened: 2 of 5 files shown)


bomshteyn:
This is great idea. 
I think though `make:table` is very generic and should be a bit more specific.

iAmKevinMcKee:
> This is great idea.
> I think though `make:table` is very generic and should be a bit more specific.

The maintainer requested this as the command. It doesn't conflict with anything else. https://github.com/rappasoft/laravel-livewire-tables/discussions/399#discussioncomment-1001762

rappasoft:
Probably a valid point though I can't imagine people installing two table plugins in one project?

iAmKevinMcKee:
A really simple upgrade to this would be to pass in an optional parameter for the model you are working with, and have that prepopulate in the `query()` method.

For example, `php artisan make:table UsersTable --model=User` would also add this:

```php
public function query(): Builder
{
    return User::query(); // this line here
}
```

It could be simplified even more by just accepting the model name as the second parameter, so

`php artisan make:table UsersTable User`

Happy to add this if you like.

iAmKevinMcKee:
Okay, I added the model as a second parameter (optional). Now you can do this:

scaffold without the model
```php artisan make:model UsersTable``` 

scaffold with User model
```php artisan make:model UsersTable User``` 

I built the logic to be smart enough to import the model class as long as the model is either in the app folder or in the app/models folder. If it's somewhere else, then I show an error in the console and just import the non-existent app/models file.

**Comment by rappasoft** [comment]
Awesome good work. I'll try to get this in this week I've just been swamped.

**Comment by iAmKevinMcKee** [comment]
I added one more feature. If you pass in `--view` or `-V` then the command will generate a row view for you in `resources/views/livewire-tables/rows/` with the following stub: 

```blade
<x-livewire-tables::table.cell>
{{--    Note: This is a tailwind cell    --}}
{{--    For bootstrap 4, use <x-livewire-tables::bs4.table.cell>    --}}
{{--    For bootstrap 5, use <x-livewire-tables::bs5.table.cell>    --}}
</x-livewire-tables::table.cell>
```

and it will also add the row view method to the class:

```php
public function rowView(): string
{
    return 'livewire-tables/rows/snake_of_classname.blade.php';
}
```

iAmKevinMcKee:
Sorry for all the commits. I'm just now getting into some good testing and I'm finding issues. I think I have worked everything out now. Feel free to squash commits.

src/LaravelLivewireTablesServiceProvider.php
```diff
@@ -32,6 +33,17 @@ public function bootingPackage(): void
         Blade::component('livewire-tables::bootstrap-5.components.table.heading', 'livewire-tables::bs5.table.heading');
         Blade::component('livewire-tables::bootstrap-5.components.table.row', 'livewire-tables::bs5.table.row');
         Blade::component('livewire-tables::bootstrap-5.components.table.cell', 'livewire-tables::bs5.table.cell');
+
+        $this->registerCommands();
```
lloricode, line 37: hi @iAmKevinMcKee please refer to this used tool https://github.com/spatie/laravel-package-tools#registering-commands

lloricode (changes requested):

lloricode:
> This is great idea.
> I think though `make:table` is very generic and should be a bit more specific.

yes, it will become confusing on large projects

iAmKevinMcKee:
Just checking in here. Anything else you'd like me to consider adding or anything I can explain?

rappasoft:
> Just checking in here. Anything else you'd like me to consider adding or anything I can explain?

Sorry I've been swamped. I'm looking at it now (though its 1am here) and will probably finish looking tomorrow. I'm going to change it to `make:datatable` just in case we conflict with something else.

I'll make a release even if this is the only thing just so people can use it.

PR #415, opened 2021-07-22
Feature/pluralization

macbookandrew:
This fixes the case where there is only one row/result for English.

### DIFF
--- a/resources/views/bootstrap-4/includes/bulk-select-row.blade.php
+++ b/resources/views/bootstrap-4/includes/bulk-select-row.blade.php
@@ -23,7 +23,7 @@
                     <span>
                         @lang('You are currently selecting all')
                         <strong>{{ number_format($rows->total()) }}</strong>
-                        @lang('rows').
+                        {{ trans_choice('rows', $rows->total()) }}
                     </span>
 
                     <button
--- a/resources/views/bootstrap-4/includes/bulk-select-row.blade.php
+++ b/resources/views/bootstrap-4/includes/bulk-select-row.blade.php
@@ -23,7 +23,7 @@
                     <span>
                         @lang('You are currently selecting all')
                         <strong>{{ number_format($rows->total()) }}</strong>
-                        {{ trans_choice('rows', $rows->total()) }}
+                        @choice('rows', $rows->total())
                     </span>
 
                     <button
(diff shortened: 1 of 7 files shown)


resources/views/bootstrap-4/includes/bulk-select-row.blade.php
```diff
@@ -23,7 +23,7 @@ class="btn btn-primary btn-sm"
                     <span>
                         @lang('You are currently selecting all')
                         <strong>{{ number_format($rows->total()) }}</strong>
-                        @lang('rows').
```
rappasoft, line 26: Wouldn't this be 'rows' regardless?

resources/views/tailwind/includes/pagination.blade.php
```diff
@@ -1,12 +1,20 @@
 @if ($showPagination)
     <div class="p-6 md:p-0">
         @if ($paginationEnabled && $rows->lastPage() > 1)
-            {{ $rows->links() }}
```
rappasoft, line 4: Why are we replacing this?

resources/views/bootstrap-4/includes/bulk-select-row.blade.php
rappasoft: I also feel like [these](https://github.com/rappasoft/laravel-livewire-tables/pull/415/files#diff-f7f9c78938d95da7404fbfe23bc215ae70b57ce1b0eb3bcc3a481e7f3716879dR9) if anything should have been replaced, the rest of the changes don't seem too worth it to me.

rappasoft:
Needs work, i'll handle.

resources/views/bootstrap-4/includes/bulk-select-row.blade.php
andrewminion-luminfire: Oh, I guess this is only used if there’s more than one page.

resources/views/tailwind/includes/pagination.blade.php
andrewminion-luminfire: The tailwind paginator lacks pluralization support.

andrewminion-luminfire (commented):
Ok; thanks!

macbookandrew:
Sorry…commented with my work account instead of personal account.

PR #436, opened 2021-08-15
Add checked to a user column selection

Majkie:
Add checked to a user column selection.

To reduce large datatables it should be possible to display only selected fields by default and leave rest hidden unless user wants to display them.

```php
Column::make('Name')->selected(),

Column::make('Created At'),
```

This will only show Name column meanwhile Created At stays hidden but still possible to display in columns dropdown.

### DIFF
--- a/src/Traits/WithColumnSelect.php
+++ b/src/Traits/WithColumnSelect.php
@@ -12,6 +12,8 @@ trait WithColumnSelect
     public bool $columnSelect = false;
     public array $columnSelectEnabled = [];
 
+    public bool $usesSelect = false;
+
     public function mountWithColumnSelect(): void
     {
         // If the column select is off, make sure to clear the session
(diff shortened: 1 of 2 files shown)


vmosoti:
Really nice feature to have

rappasoft:
Would this not be a breaking change? It would cause everyone's columns to be hidden unless they added that method call.

Majkie:
Didn't think this far. I'll try to test and fix it.

rappasoft:
The shown cols are also stored in a session so it might get hairy on what's stored in the session vs the code, etc.

Majkie:
I can't think of anything but adding something like `public bool $usesSelected = false;` which would be set to true if even one of many columns had selected(). And based on value of `usesSelected`, add $column->isSelected() to the collection filter.

rappasoft:
Could work, this really only matters for the first time the table is loaded for the other. Because it saves it to the session and is then loaded from that every other time.

Issue #381, opened 2021-06-29
Allow column filter callbacks to rename their parameter


natepisarski:
(**Originally posted as a discussion accidentally - feel free to delete.** It got posted there because I marked it 'Feature' instead of bug since this is technically working as expected.

<div type='discussions-op-text'>

<sup>Originally posted by **natepisarski** June 29, 2021</sup>
This request is for letting users of this package rename the parameter in the Filter callback for columns.

The root cause of the issue appears to be [this line](https://github.com/rappasoft/laravel-livewire-tables/blob/1c83a63e0670ea0705021a73f07af5d84d0d8f0c/src/Views/Column.php#L238) in the Column class.

# Example of column that works

```php
 Column::make('Thing', 'thing')
            ->format(fn ($value) => view('components.show-thing', ['thing' => $value]))
            ->asHtml()
```

# Example of column that doesn't work
```php
 Column::make('Thing', 'thing')
            ->format(fn ($thing) => view('components.show-thing', ['thing' => $thing]))
            ->asHtml()
```

This would fail with an error such as:

```
Unable to resolve dependency [Parameter #0 [ <required> $thing ]] in class  ...
```

# Rationale
This is overall a very minor issue. But it prevents the use of `compact`. So in the above non-working example, the formatting line could be shortened to:

```php
->format(fn ($thing) => view('components.show-thing', compact('thing'))
```
if it didn't need to be named `value`.</div>

rappasoft:
What's your recommendation for a fix?

natepisarski:
Sorry for the world's latest response. I don't have a 100% great idea for a fix. But I've been thinking about this more recently, and was wondering what something like (pseudocode) would be like:

```php
if ($this->formatCallback) {
  $callbackParameters =  (new \ReflectionFunction($callback))->getParameters();
  if ( /* callback parameters has 'value', 'column', or 'row' */) {
      return app()->call($this->formatCallback, ['value' => $value, 'column' => $column, 'row' => $row]);
  } else {
    // The filter callback doesn't specify a name we're familiar with; we should call it using positional arguments
    $this->formatCallback($value, $column, $row);
  }
}
```
In theory this keeps backwards compatibility, and lets people decide whether or not they want to use their format callbacks positionally or not. The isuse is that for every `formatCallback` that is processed, we need to use reflection to read the arguments. It's not a big performance hit but it goes without saying that it's less performant than what's there now.

rappasoft:
Fixed in next release.

PR #438, opened 2021-08-18
Fix format callback

Majkie:
Related to Issue #381 

Now this thing works properly and does not require parameter containing row value to be called `$value`.

```php
Column::make('Thing')->format(fn ($thing) => view('components.show-thing', compact('thing'))
```

### DIFF
--- a/src/Views/Column.php
+++ b/src/Views/Column.php
@@ -260,7 +260,7 @@ class Column
         $value = data_get($row, $columnName);
 
         if ($this->formatCallback) {
-            return app()->call($this->formatCallback, ['value' => $value, 'column' => $column, 'row' => $row]);
+            $value = call_user_func($this->formatCallback, $value, $row, $column);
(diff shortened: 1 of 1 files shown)


rappasoft:
They are not in the correct order but I'll change it.

decantr:
I might be missing something but does this change actually improve anything?

From my understanding, we now have to pass dummy arguments in place where we would usually just need one named argument before, for example accessing the `$row` property?

OmarAbdelSamea:
I was using version 1.12 and after updating to 1.14 all tables with custom format failed because I was passing only `$row` it returned error that `$row` is null.

```php
 ->format(function ($row) {
          return view('exampleDir.exampleFile', ['row' => $row]);
  }),
```
I had to pass dummy parameters to make it work.
```php
 ->format(function ($value, $column, $row) { ... })
```

Issue #412, opened 2021-07-22
Problem with bulk select selectedRowsQuery & selectedKeys


KevinYobeth:
When I run bulk action while selecting more than 1 row using search without clearing the search field, `selectedRowsQuery` & `selectedKeys` will only return 1 row instead of all the selected row. If I clear the search field first and than does the bulk action, it works propperly.

![image](https://user-images.githubusercontent.com/10485985/126681164-138137b4-3adc-40ff-8866-7ba12568ffeb.png)

![image](https://user-images.githubusercontent.com/10485985/126681403-2435716f-33bc-4569-a724-d5581ae83c48.png)

PR #439, opened 2021-08-18
Fix bulk select with search and add typed property to selected

Majkie:
#412

### DIFF
--- a/src/Traits/WithBulkActions.php
+++ b/src/Traits/WithBulkActions.php
@@ -67,10 +67,12 @@ trait WithBulkActions
     /**
      * @return Builder|Relation
      */
-    public function selectedRowsQuery()
+    public function selectedRowsQuery(): Relation|Builder
     {
-        return (clone $this->rowsQuery())
-            ->unless($this->selectAll, fn ($query) => $query->whereIn($query->qualifyColumn($this->primaryKey), $this->selected));
+        return $this->query()->unless(
+            $this->selectAll,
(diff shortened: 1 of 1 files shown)


rappasoft:
This is a small change but I don't know what it's going to do to the rest of the functionality. There is extra functionality in the rowsQuery method that is now being ignored.

Majkie:
There shouldn't be any problem because it takes query passed down by user without any other parameters, filters etc. and affects only selected rows based on array of selected ids. rowsQuery is still used but not for bulk select action. Also I tested it a little and didn't notice any problem.

rappasoft:
How do I set it up to get the error that this fixes?

Majkie:
Select 1 column and then search for some other column to select it so the previous one is not present in table. Then if you want to run some action you only get searched result because it clones query with search parameters. Fix is just simply taking base query provided by user without any changes.

rappasoft:
Having a hard time reproducing this issue but if it fixes #412 I want to figure it out.

KevinYobeth:
**Steps to Reproduce:**
1. Select 1 column
2. Search for any other entries and select 1 of them
3. Run bulk action

Result: Only the column that is being searched is on the `selectedRowsQuery` and `selectedKeys`
Expected: Both the searched and previously selected column should be on `selectedRowsQuery` and `selectedKeys`

https://user-images.githubusercontent.com/10485985/136682253-737cbef8-401a-41de-a5b3-b8d4870ce50f.mov

PR #441, opened 2021-08-19
Added multiselect filter

Ypho:
I've added a multiselect option filter.

### DIFF
--- a/resources/views/bootstrap-4/includes/filter-pills.blade.php
+++ b/resources/views/bootstrap-4/includes/filter-pills.blade.php
@@ -4,14 +4,23 @@
             <small>@lang('Applied Filters'):</small>
 
             @foreach($filters as $key => $value)
-                @if ($key !== 'search' && strlen($value))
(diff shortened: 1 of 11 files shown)


Majkie:
I'm not exactly sure if its only problem on my side that should be documented, but whenever I want to uncheck one of many checked checkboxes it unchecks all of them.

Ypho:
@Majkie got the same. I thought I fixed it all when I moved from multiselectbox to checkboxes, but I didn't see this. Will try and fix it when I got some time.

rappasoft:
If I dump the values it only returns an array of the most previously selected, even if there are multiple selected.

rappasoft:
I've pumped many hours into trying to get this to work in all scenarios. I have yet to get it to work with both string and numeric array keys. If anyone else has time feel free.

juliomotol:
@rappasoft why was this closed?

rappasoft:
It doesn't work and has been stale.

rappasoft:
Yes, the one here was almost there but it was buggy. Also please modify all 3 templates.

PR #458, opened 2021-09-17
Allow to disable responsive status of the table

fabio-ivona:
While developing for a new project I faced an issue while building a table with some long cell content that I wish to wrap instead of making my table to scroll:

image

My wish is to be able to disable the table responsiveness and have full control over cells class with the given methods. In tailwind setup this seems to possible to me.

So I made this PR in order to add a new property to the DataTableComponent:

```PHP
class Table extends DataTableComponent
{
    use ConfirmsActions;

    public bool $responsive = false;   //<--- setting this to 'false' would disable table responsiveness

    public string $defaultSortColumn = 'created_at';
    public string $defaultSortDirection = 'desc';

    //...
```

Obviously the `$responsive` property is **set to `true` by default**, in order to avoid breaking changes

@rappasoft let me know what do you think about it!


fabio-ivona:
@rappasoft tests are failing, but it not seems related to my implementation, they fail on master branch as well

fabio-ivona:
Merged new changes from master

PR #461, opened 2021-09-21
Add linkTo cell setting

fabio-ivona:
Hi @rappasoft 

as usual: you're doing a great work! A lot of livewire datatables packages are showing up, but yours is the one that better balances features and DX!

I'll try to contribute by adding a new `->linkTo(...)` method to a column:

with this PR, during the table columns setup you could do:

```PHP
return [
            //...

            Column::make('Name', 'name')
                ->linkTo(fn($value, $column, $row) => route('pets.edit', $row->id))
                ->searchable(),

            //...
        ];
```

and this will result in the table cell being rendered with this content:

`<a href='http://my.domain/pets/5/edit'>My Pet 5 Name</a>`

So the cell would be clickable and be linked to the url generated by the `->linkTo()` callback:

**Note:** a second parameter for the `->linkTo()` method allows to hint the link target:

```PHP
  Column::make('Name', 'name')
                ->linkTo(fn($value, $column, $row) => route('pets.edit', $row->id), '_blank')
                ->searchable(),
```

What do you think?


rappasoft:
When would this be more useful than this though? https://rappasoft.github.io/laravel-livewire-tables-docs/rows/Clickable-Rows

fabio-ivona:
This is intended to work on a single cell, a different use case

And, as it would be defined in the Column

fabio-ivona:
This would work on a single cell, so a different use case

kevyworks:
Can we merge this? If the library has like Macro I would go with that approach as well.

PR #467, opened 2021-09-24
Add bulk actions function

fabio-ivona:
hi @rappasoft!

this PR would add the capability to define bulkActions inside a function, this wil allow to generate them dynamically (for example using translation functions) or switching actions based on context

the old way of defining bulkActions will remain enabled:

```php
class Table extends DataTableComponent
{
    //...
    public array $bulkActions = [  ... ]

```

but with this PR you could write this, as an alternative:

```php
class Table extends DataTableComponent
{
    //...
    public function bulkActions(): array
    {
        return [
              'count' => __('pets.count'),
        ]
    }
```

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -2,6 +2,7 @@ name: run-tests
 
 on: [push, pull_request]
 
+
 jobs:
   test:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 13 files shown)


rappasoft:
Thank you for both modifying all 3 templates as well as adding tests!

fabio-ivona:
You're welcome!

PR #474, opened 2021-09-30
add/"themeColor"-Variable to set the color

Enaah:
Hi mate,

first, thanks for this great package and the quick development 🙌🏼

What does this PR do:
It allows the user to easily (without modifying the published views) change the _themecolor name_ (`indigo`).

I wanted to adjust the theming a bit and replaced the colors in the views. Because of the fast development I had to republish these several times (to make use of the new features) and my color adjustments were gone.

⚠️ Because the colors are generated _dynamically_ [**purgecss** from Tailwind](https://tailwindcss.com/docs/optimizing-for-production) will not work with these colors.
Maybe there should be a warning to add these colors to allow list:
* bg-[color]-50
* bg-[color]-100
* bg-[color]-200
* bg-[color]-300
* ring-[color]-100
* border-[color]-300
* text-[color]-600
* text-[color]-800
* bg-[color]-900
* text-[color]-400
* text-[color]-500

```js
// tailwind.config.js
module.exports = {
  purge: {
    content: ['./src/**/*.html'],
    safelist: [
     'bg-[color]-50',
    'bg-[color]-100',
    'bg-[color]-200',
    'bg-[color]-300',
    'ring-[color]-100',
    'border-[color]-300',
    'text-[color]-600',
    'text-[color]-800',
    'bg-[color]-900',
    'text-[color]-400',
    'text-[color]-500'
      // ...
      'lg:text-right',
    ]
  },
  // ...
}
```

Changes
---
- replaces the `indigo`-color with the `themeColor`-Property
- replaces every color with the new dynamic color-name

### DIFF
--- a/resources/views/tailwind/includes/bulk-actions.blade.php
+++ b/resources/views/tailwind/includes/bulk-actions.blade.php
@@ -11,7 +11,7 @@
                     <button
                         x-on:click="open = !open"
                         type="button"
-                        class="inline-flex justify-center w-full rounded-md border border-gray-300 shadow-sm px-4 py-2 bg-white text-sm font-medium text-gray-700 hover:bg-gray-50 focus:border-indigo-300 focus:ring focus:ring-indigo-200 focus:ring-opacity-50 dark:bg-gray-700 dark:text-white dark:border-gray-600 dark:hover:bg-gray-600"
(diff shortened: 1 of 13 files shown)


rappasoft:
The whole reason I did not do this was because of the purging issue. I wouldn't know any other way to do it other than a custom theme, but I don't know if I can switch it now without it being a breaking change.

PR #481, opened 2021-10-06
Add Bootstrap Clases to Filter

nessimabadi:
Adds a class for the Bootstrap 4 and 5 themes for the dropdown Filter Button.
Currently it only has the main class 'btn' which doesn't contain any styling.
This PR adds a styling class 'btn-outline-secondary' which is the same one as the Clear Search button on the search component.

### DIFF
--- a/resources/views/bootstrap-5/includes/filters.blade.php
+++ b/resources/views/bootstrap-5/includes/filters.blade.php
@@ -9,7 +9,7 @@
         <button
             x-on:click="open = !open"
             type="button"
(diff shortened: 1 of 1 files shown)


rappasoft:
With this, it makes it less streamlined because one looks different doesn't it?.

Before:

Screen Shot 2021-10-11 at 23 15 51
Screen Shot 2021-10-11 at 23 15 45

After:

Screen Shot 2021-10-11 at 23 17 18
Screen Shot 2021-10-11 at 23 17 13

PR #492, opened 2021-10-14
add row number column to tailwind theme

ratiw:
Enable `tailwind` table to show row number by setting 
```php
    $rowNumberEnabled = true;
```

### DIFF
--- a/resources/views/tailwind/includes/table.blade.php
+++ b/resources/views/tailwind/includes/table.blade.php
@@ -21,6 +21,12 @@
             </x-livewire-tables::table.heading>
         @endif
 
+        @if ($rowNumberEnabled)
+            <x-livewire-tables::table.heading>
+                #
+            </x-livewire-tables::table.heading>
+        @endif
+
         @foreach($columns as $column)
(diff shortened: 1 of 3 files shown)


rappasoft:
What is the use case for this?

rappasoft:
Yes but when you sorted or filtered the row number associated with the rows would change. So you wouldn't be able to say "row with number 5 is the same all the time" so I could see issues. That's why I like having a `sort` column in the demo, its the same number regardless of how its sorted or filtered.

Unless you see some other use I don't see.

ratiw:
Hmm, am I correct that you assumed that the "row number" is the "record ID"? 

If so, that is unintentionally misunderstanding for my implementation. To me, the "row number" is like a line number in the table. It wasn't meant to be locked to any particular record at all. Sorry for the misunderstanding.

rappasoft:
I think it will add a lot of unnecessary code since you need to make it work with the header, secondary header, footer, reordering, etc. Plus you only did the tailwind theme and you would have to do it for bootstrap 4 and 5 as well.

PR #495, opened 2021-10-14
 add custom class to table

CristhoferMF:
code-snapshot


CristhoferMF:
Sorry. That bug escaped me
Now it was tested. 
All was ok.
... Mi first pull request :+1:

PR #535, opened 2021-10-28
fix bulk actions breaking change

fabio-ivona:
@rappasoft  this PR will fix issues in project with custom views after adding the `bulkActions()` function support see (#467 and #517)

custom views that use old `$bulkActions` notation instead of `$this->bulkActions` (needed for computed property) are breaking (I think this is why we ended adding and removing $bulkActions property multiple times)

for explanation of the issue, see #533 

the fix is to pass a `$bulkActions` attribute to the rendered view in order to allow the old notation to work

this can be eventually removed in a future major version

### DIFF
--- a/src/DataTableComponent.php
+++ b/src/DataTableComponent.php
@@ -223,6 +223,7 @@ abstract class DataTableComponent extends Component
                 'customFilters' => $this->filters(),
                 'rows' => $this->rows,
                 'modalsView' => $this->modalsView(),
+                'bulkActions' => $this->bulkActions
             ]);
     }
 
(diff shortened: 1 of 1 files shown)


aeq-dev:
Hello there,
The bulk action drop-down doesn't work if I select some records and do some filters.
Bootstrap 5

fabio-ivona:
@aeq-dev have you customized any view?

aeq-dev:
@fabio-ivona never mind, sorry I did update for package but didn't republish views.
Thanks

fabio-ivona:
There's no need to publish views if you don't have to change them, it is discouraged in the docs, also 😉

aeq-dev:
Yes I know, I did some changes in row, filters and bulk action :)

PR #585, opened 2021-12-02
Added Datetime Filter

Wit3:
Added datetime-local input filter

[https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local
)


rappasoft:
Bump

Issue #625, opened 2022-01-25
bulk action selection does not preserve search filter


Lupennat:
how to replicate the bug:

- no filter applied total records = 100
- filter results through search query, total records = 50
- select checkbox in the table head
- click select all (the number of selected elements respects the maximum number of filtered records 50)
- execute a bulk action
- the action is executed on all 100 records instead of 50 selected

possible solution:

```php
    /**
     * @return Builder|Relation
     */
    public function selectedRowsQuery()
    {
        $query = $this->query();

        return $query->when($this->selectAll,
            fn ($query) => $this->applySearchFilter($query),
            fn ($query) => $query->whereIn($query->qualifyColumn($this->primaryKey), $this->selected)
        );
    }
 ```

bureauvk:
Can you show us your bulk method? Are you only applying changes to the selectedkeys?
I cant replicate this issue.

Lupennat:
you can check it directly on the demo page https://tables.laravel-boilerplate.com/tailwind.
Type something in the search that reduces the number of users but has more than one page (like "er"), then click the checkbox next to "sort" and then "select all" on the bar. 
You will see X items selected, apply the "deactivate" bulk action, remove the search filter and you will see that all users are disabled instead of X.

PR #632, opened 2022-02-02
fix select rows query

Lupennat:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

more info on issue https://github.com/rappasoft/laravel-livewire-tables/issues/625

### DIFF
--- a/src/Traits/WithBulkActions.php
+++ b/src/Traits/WithBulkActions.php
@@ -68,8 +68,8 @@ trait WithBulkActions
      */
     public function selectedRowsQuery()
     {
-        return $this->query()->unless(
-            $this->selectAll,
+        return $this->query()->when($this->selectAll,
+            fn ($query) => $this->applySearchFilter($query),
             fn ($query) => $query->whereIn($query->qualifyColumn($this->primaryKey), $this->selected)
         );
     }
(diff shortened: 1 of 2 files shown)


Lupennat:
@rappasoft any chance to approve this pull request?

affektde:
I think this introduced a bug, when we use columnSearch filter. (https://rappasoft.com/docs/laravel-livewire-tables/v1/columns/column-header-search)

`$this->query()` holds already a search query from a possible columnSearch.

The resulting `wheres` with `and` `or` `and` queries all items with with `title` `%lo%"` in my example.

`selectedRowsQuery` then holds the whole page instead of the one selected item.(id:4046 in my example) 

image

I'll try to find a solution for that, but maybe someone has a quick fix for it.

affektde:
Okay sorry for the noise.

For those who may have the same problem: don't use `orWhere` in your `columnSearch` queries.

image

Lupennat:
Hi @affektde PR, only change behaviour when "select All" Checkbox is enabled. Are you sure the problem is related to the pull request?

Lupennat:
@affektde you can use where with closure like this:

```php
$query->where(function($query) {
    $query->where('title', 'like', '%' . $value . '%')
          ->orWhere('version', 'like', '%' . $value . '%');
});
```

PR #643, opened 2022-02-07
Fix: not showing dropdowns after sort column

vildanbina:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/resources/views/bootstrap-4/includes/column-select.blade.php
+++ b/resources/views/bootstrap-4/includes/column-select.blade.php
@@ -10,7 +10,6 @@
                 x-on:click="open = !open"
                 class="btn dropdown-toggle d-block w-100 d-md-inline"
                 type="button"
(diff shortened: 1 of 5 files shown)


bureauvk:
Your fix works, but why? You just deleted some aria-labelledby attributes and id attributes.

vildanbina:
> Your fix works, but why? You just deleted some aria-labelledby attributes and id attributes.

after dropdown replaced by livewire response, I don't know why alpine js version 3 has crashed from id attribute

rappasoft:
Just because it works doesn't mean it's a viable fix. We need to understand why it's behaving that way so we can properly fix it.

bureauvk:
> Going to merge for now but it's not a permanent fix.

I don't know if it's worth mentioning, but in the demo it seems to work fine. Or is the demo running on the V2?

PR #677, opened 2022-03-25
Add Actions and UI refactor

trippo:
### Add actions functionality to render buttons on the top right corner of the table
  
Buttons works like bulk actions, you can add an array with buttons to include:
```
public array $actions = [
        'createResource' => 'Action Title',
    ];
```
image

### Add bulkActionsInline option to render bulk actions as inline buttons instead of dropdown
You can define a bool attribute to enable this option
`public bool $bulkActionsInline = true;`

image

### Select filter default label 
You can use the second params of the select function of filters to set the label of the default Option (the option with the '' value, selected when the filter is empty).
image

### Move per page dropdown on the bottom left part of the table
image

- use default Laravel pagination

## Note 
I changed only Tailwind template and I haven't updated your doc yet.
If do you like my change I can do it later

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

* [ ] Have you successfully run tests with your changes locally?


rappasoft:
This is a generous PR, but with V2 about to come out I don't have time to test this and build it for the other templates. We can revisit for V2. Thanks.

PR #702, opened 2022-04-06
Allow query for other column types

bumbummen99:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

In case the ImageColumn or LinkColumn is **not** supposed to use the primary key of the row ImageColumn and LinkColumn can not be used right now, their field value is completely irrelevant. Added a check for field so that it will be queried too if set (why else would you set it for these columns):
```php
            (new ImageColumn('Image', 'image'))
                ->location(function($row) {
                    return $row->image;
                })
                ->attributes(function($row) {
                    return [
                        'class' => 'w-8 h-8 rounded-full',
                    ];
                })
```

### DIFF
--- a/docs/columns/other-column-types.md
+++ b/docs/columns/other-column-types.md
@@ -79,6 +79,11 @@ ImageColumn::make('Avatar')
     ->location(function($row) {
         return storage_path('app/public/avatars/' . $row->id . '.jpg');
     })
+// Or if you want to query a column instead of using primary key
+ImageColumn::make('Avatar', 'avatar')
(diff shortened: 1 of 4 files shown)


rappasoft:
Can I have an example of why this is needed?

bumbummen99:
> Can I have an example of why this is needed?

My image can be a local file, an s3 temporary url or just any http* so just using ID does not work for me here. Actually the image property is an `\Illuminate\Database\Eloquent\Casts\Attribute`. I fell back to creating my own column type before moving to another solution.

PR #729, opened 2022-04-19
Adding NumberFilter

cubemastercodes:
Greetings,

Hope you are doing great...

Thank you a lot for this great package. 

While we are using this package, we've discovered a need of having a numeric filter i.e. quantity more than, quantity less than. 

below is one of the usages of NumberFilter

image

and this is the filter configuration:

image

Regards,


gpibarra:
can you add TextFilter?

PR #734, opened 2022-04-22
SQL quotes replaced to support Postgresql

marcheffels:
Hey,

tested with MySQL and PostgreSQL.

Greeting Marc

### DIFF
--- a/src/Traits/WithSorting.php
+++ b/src/Traits/WithSorting.php
@@ -80,7 +80,7 @@ trait WithSorting
             } elseif ($column->isBaseColumn()) {
                 $builder->orderBy($column->getColumnSelectName(), $direction);
             } else {
-                $builder->orderByRaw('`'.$column->getColumnSelectName().'`' . ' ' . $direction);
+                $builder->orderByRaw('"'.$column->getColumnSelectName().'"' . ' ' . $direction);
             }
         }
 


rappasoft:
Are these quotes not working?

marcheffels:
> Are these quotes not working?

It worked for MySQL but there is a SQL syntax error for PostgreSQL. With this quotes it will works for both.

ddejong98:
Hi,

Unfortunately, this commit made that sorting in my application is not working anymore. I am using standard Laravel Sail for my local environment (comes with MySQL 8.0).

The only way I can make it work is by reverting the change in this commit. If there are some other things I could try, please let me know.

With kind regards,
Damian

ddejong98:
> Are you getting any errors?

I don't get any errors, it just won't sort on a nested column. Perhaps an example can clearify the issue I am running into. Below is the query that I got from the function `$this->getQuerySql()`:

`select `models`.`id` as `id`, `article_types`.`name` as `article_type.name`, `brands`.`name` as `brand.name`, `models`.`name` as `name`, `models`.`code` as `code`, `models`.`year` as `year`, `models`.`colors` as `colors`, `countries`.`name` as `country.name`, `models`.`created_at` as `created_at`, `models`.`updated_at` as `updated_at` from `models` left join `article_types` on `models`.`article_type_id` = `article_types`.`id` left join `brands` on `models`.`brand_id` = `brands`.`id` left join `countries` on `models`.`country_id` = `countries`.`id` where `active` = true and `models`.`deleted_at` is null order by "brand.name" asc, `year` desc, `code` asc`

The result, also when I run the query directly in MySQL, is that the data is only sorted on the second and third columns in the `ORDER BY` clause. When I change the double quotes around `brand.name` to backticks, the result gets sorted properly.

It thought it might have to do with the fact that MySQL does not necesarily treat double quotes the same as backticks. As far as I understand correctly, it only sees a string enclosed in double quotes as an identifier in case the SQL mode `ANSI_QUOTES` is enabled (see: https://dev.mysql.com/doc/refman/8.0/en/string-literals.html). However, after having enabled that mode, I still get the same result.

Issue #721, opened 2022-04-15
Filter Drop Down open after sorting


AdamKyle:
I am not sure if this is a me issue or a library issue but heres how to replicate:

I created a component: `admin.items.item-table` which looks like:

```
<?php

namespace App\Flare\View\Livewire\Admin\Items;

use Illuminate\Database\Eloquent\Builder;
use Rappasoft\LaravelLivewireTables\DataTableComponent;
use Rappasoft\LaravelLivewireTables\Views\Column;
use App\Flare\Models\Item;
use Rappasoft\LaravelLivewireTables\Views\Filters\SelectFilter;

class ItemsTable extends DataTableComponent {

    public function configure(): void {
        $this->setPrimaryKey('id');
    }

    public function builder(): Builder {
        return Item::whereNotIn('type', ['quest', 'alchemy'])
            ->whereNull('item_prefix_id')
            ->whereNull('item_suffix_id')
            ->when($this->getAppliedFilterWithValue('types'), fn ($query, $type) => $query->where('type', $type));
    }

    public function filters(): array {
        return [
            SelectFilter::make('Types')
                ->options([
                    'weapon'        => 'Weapons',
                    'bow'           => 'Bows',
                    'body'          => 'Body',
                    'helmet'        => 'Helmets',
                    'shield'        => 'Shields',
                    'sleeves'       => 'Sleeves',
                    'gloves'        => 'Gloves',
                    'leggings'      => 'Leggings',
                    'feet'          => 'Feet',
                    'ring'          => 'Rings',
                    'artifact'      => 'Artifacts',
                    'spell-healing' => 'Healing Spells',
                    'spell-damage'  => 'Damage Spells',
                ]),
        ];
    }

    public function columns(): array {
        return[
            Column::make('Name')->searchable(),
            Column::make('Type')->searchable()->format(function ($value) {
                return ucfirst(str_replace('-', ' ', $value));
            }),
            Column::make('Min Crafting Lv.', 'skill_level_required')->sortable(),
            Column::make('Trivial Crafting Lv.', 'skill_level_trivial')->sortable(),
            Column::make('Cost')->sortable()->format(function ($value) {
                return number_format($value);
            }),
        ];
    }
}
```

This table loads fine, and I can filter fine. But if I then attempt to sort the filter drop down will open. Clicking away, will not close the filter drop down.

I have not made any changes to the view files other then adding some slight padding to the search element.

If I filter, with out sorting it works fine.

[...]

PR #740, opened 2022-04-26
Fixes Filter Freeze/Open when a Column is Sorted

cubemastercodes:
This should fix [Filter Drop Down open after sorting](https://github.com/rappasoft/laravel-livewire-tables/issues/721#issue-1205935094)


cubemastercodes:
@rappasoft 😂 I don't know why... but this fixes it

rappasoft:
Alright i'll accept it for now but we really need to understand why lol

PR #757, opened 2022-05-04
Change formatting of relational column to allow proper sorting.

korki696:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

When sorting a relational column, the table wouldn't actually sort. Changing the query to use a backquote (`) instead of a double quote (") seems to fix this.

### DIFF
--- a/src/Traits/WithSorting.php
+++ b/src/Traits/WithSorting.php
@@ -80,7 +80,7 @@ trait WithSorting
             } elseif ($column->isBaseColumn()) {
                 $this->setBuilder($this->getBuilder()->orderBy($column->getColumnSelectName(), $direction));
             } else {
-                $this->setBuilder($this->getBuilder()->orderByRaw('"'.$column->getColumnSelectName().'"' . ' ' . $direction));
+                $this->setBuilder($this->getBuilder()->orderByRaw('`'.$column->getColumnSelectName().'`' . ' ' . $direction));
             }
         }
 


rappasoft:
We need a way for this to both work in MySQL and postgres. Personally, I can't replicate this though.

korki696:
I am running on MySQL8 and even just a raw query will only work with backquotes. I don't have a Postgres spun up or am to familiar with it. Are you able to sort a relational column on MySQL with the current implementation?

rappasoft:
Should we do a conditional and just set it differently for the different drivers?

mitulgolakiya:
Facing the same issue. I had a column with relationship sorting. Everything was working fine for version 2.1.1. After upgrading to 2.2.0+, none of the related columns are working.

Seems like the problem is when order by column name is replaced with double quotes `"` instead of `, then sorting is stopped with MySQL.

vishalinfyom:
Yes, facing same issue.

 after upgrading to latest version in MySQL, sorting is not working.

bskiefer:
I was able to resolve by adding this method to my controllers:

```php
    public function getRows()
    {
        $this->baseQuery();
        $query = $this->getBuilder()->getQuery();
        $orders = $query->orders;
        $query->orders = [];

        foreach ($orders as $order) {
            $query->orders[] = [
                'type' => 'Raw',
                'sql' => str_replace('"', '`', $order['sql'])
            ];
        }
        $this->getBuilder()->setQuery($query);

        return $this->executeQuery();
    }
```

rappasoft:
Closing this in favor of this fix: https://github.com/ddejong98/laravel-livewire-tables/commit/ad96a80bcbc215ab06a41ee5a8eff02424301758

PR #764, opened 2022-05-07
Column Filters

rappasoft:


inmanturbo:
looks like we have menus plural in one method and singular in the other.

```php
    /**
     * @return bool
     */
    public function isHiddenFromMenus(): bool
    {
        return $this->hiddenFromMenus === true;
    }

    /**
     * @return bool
     */
    public function isVisibleInMenu(): bool
    {
        return $this->hiddenFromMenus === false;
    }
```

The test is expecting plural

```php
    /** @test */
    public function can_check_if_filter_is_hidden_from_menus(): void
    {
        $filter = SelectFilter::make('Active');

        $this->assertFalse($filter->isHiddenFromMenus());
        $this->assertTrue($filter->isVisibleInMenus());

        $filter->hiddenFromMenus();

        $this->assertTrue($filter->isHiddenFromMenus());
        $this->assertFalse($filter->isVisibleInMenus());
    }
```

inmanturbo:
the view is using the singular
@rappasoft I was going to fix this but I'm not sure which you want to use, singular or plural?
```blade
      @foreach($component->getFilters() as $filter)
          @if($filter->isVisibleInMenu())
              <div class="py-1" role="none">
                  <div class="block px-4 py-2 text-sm text-gray-700 space-y-1" role="menuitem">
                      <label for="{{ $component->getTableName() }}-filter-{{ $filter->getKey() }}"
                          class="block text-sm font-medium leading-5 text-gray-700 dark:text-white">
                          {{ $filter->getName() }}
                      </label>

                      {{ $filter->render($component) }}
                  </div>
              </div>
          @endif
      @endforeach
```

inmanturbo:
Also, I have notice before but it looked like most everywhere else you are using the '@continue' pattern in your loops with the inverse boolean. Might it be better to use the same pattern here?

rappasoft:
Thanks I fixed it

rappasoft:
@inmanturbo can you give it one more look over before I merge it into develop? Make sure I didn't over look anything.

inmanturbo:
> @inmanturbo can you give it one more look over before I merge it into develop? Make sure I didn't over look anything.

@rappasoft I don't see any typos or anything missing here. anything missing here Docs :white_check_mark: changelog :white_check_mark: tests :white_check_mark:  . More importantly it doesn't seem to break anything. Pretty painless edition: adds some nice features without changing the existing api. :+1:

inmanturbo:
Painless for the end user I mean. lol

PR #774, opened 2022-05-13
Update SearchHelpers.php

bobimicroweber:
This will update the state of table.. Cause when you click clear search, the old data view already exist


rappasoft:
What do you mean update the state? The table already reacts accordingly when the search is cleared.

PR #775, opened 2022-05-14
Column select

### LINKED ISSUE #760: Randomly dissappearing columns
Hi,

First off, great package!

We're using livewire tables (currently 20+ cruds) extensively thoughout a new project and are experiencing issues with columns not showing by default in the tables. Sometimes one or two columns will display, sometimes all columns will display and sometimes no columns at all will display.

I've tried adding $this->setColumnSelectStatus(true); into the configure method, but this doesn't appear to have any effect.

One possible thought is quite a few of the tables contain the same column names i.e. title, description, enabled etc and I notice that you're creating the session key by md5 hashing the column name, so I don't know whether this might be affecting things?

Any thoughts?

inmanturbo:
This feature uniquely identifies each table for column selections stored in the session. It also allows for columns selections to be put into the query string, so that selections can be bookmarked.

The tldr; is that we now have a fingerprinting method that uses `static::class` to generate a key, or slug for each component, to be used as an alias for `$tableName` in the session and query string arrays, which is always the same for the same component, and always different for another.

```php
    /**
     * returns a unique id for the table, used as an alias to identify one table from another session and query string to prevent conflicts
     */
    public function dataTableFingerprint(): string
    {
        $className = str_split(static::class);
        $crc32 = sprintf('%u', crc32(serialize($className)));

        return base_convert($crc32, 10, 36);
    }
```

```php
    /**
     * Set the custom query string array for this specific table
     *
     * @return array|\null[][]
     */
    public function queryString(): array
    {
        if ($this->queryStringIsEnabled()) {
            return [
                $this->getTableName() => ['except' => null, 'as' => $this->dataTableFingerprint()],
                'userSelectedColumns' => ['except' => null, 'as' => $this->dataTableFingerprint() . '-c']
            ];
        }

        return [];
    }
```

This produces urls that look something like this:
`http://table-demo.test/?11mbmzx[filters][year]=2022&11mbmzx-c[0]=year`

Where `11mbmzx` is the `fingerprint`, or unique `alias` for the component.

The dynamic link above will apply a filter with the key of `year` and value `2022` and select to only have the `year` column displayed.

checking only the `year` column in the columns▼ dropdown on a table fingerprinted `11mbmzx` will put `11mbmzx-c[0]=year` into the query string. it will also put the following array into the session:

 `11mbmzx-columnSelectEnabled` => array:1 [▼
    0 => `year`
  ]

Query strings will be evaluated first to determined which columns should be displayed, after that the session, followed by the default.

closes #760

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to `laravel-livewire-tables` will be documented in this file
 
 ## [Unreleased]
 
+## [dev-column-select] - 2022-05-14
+
+- Added functionality to bookmark or deep link column selection
+- Added functionality to identify different datatable components as unique in column selection and filter arrays
+
 ## [2.7.0] - 2022-05-07
 
 ### Added
--- a/docs/misc/multiple-tables.md
+++ b/docs/misc/multiple-tables.md
@@ -33,55 +33,6 @@ If you need the above, you should make them different components like so:
 <livewire:pending-users-table />
 ```
 
-## Introduction
-
-By default, your table has a name of `table`, as well as an internal array called `$table` which saves its state to the query string.
-
-The query string would look like this:
-
-```php
-// Under the hood
-public array $queryString = [
-    'table' => [
-        'search' => null,
-        'sort' => [],
-        ...
-    ],
-]
-```
-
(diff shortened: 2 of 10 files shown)


inmanturbo:
This was closed in favor of #776. Which is the same outcome with less commits.

PR #776, opened 2022-05-14
Column select

### LINKED ISSUE #760: Randomly dissappearing columns

inmanturbo:
This feature uniquely identifies each table for column selections stored in the session. It also allows for columns selections to be put into the query string, so that selections can be bookmarked.

The tldr; is that we now have a fingerprinting method that uses `static::class` to generate a key, or slug for each component, to be used as an alias for `$tableName` in the session and query string arrays, which is always the same for the same component, and always different for another.

```php
    /**
     * returns a unique id for the table, used as an alias to identify one table from another session and query string to prevent conflicts
     */
    public function dataTableFingerprint(): string
    {
        $className = str_split(static::class);
        $crc32 = sprintf('%u', crc32(serialize($className)));

        return base_convert($crc32, 10, 36);
    }
```

```php
    /**
     * Set the custom query string array for this specific table
     *
     * @return array|\null[][]
     */
    public function queryString(): array
    {
        if ($this->queryStringIsEnabled()) {
            return [
                $this->getTableName() => ['except' => null, 'as' => $this->dataTableFingerprint()],
                'userSelectedColumns' => ['except' => null, 'as' => $this->dataTableFingerprint() . '-c']
            ];
        }

        return [];
    }
```

This produces urls that look something like this:
`http://table-demo.test/?11mbmzx[filters][year]=2022&11mbmzx-c[0]=year`

Where `11mbmzx` is the `fingerprint`, or unique `alias` for the component.

The dynamic link above will apply a filter with the key of `year` and value `2022` and select to only have the `year` column displayed.

checking only the `year` column in the columns▼ dropdown on a table fingerprinted `11mbmzx` will put `11mbmzx-c[0]=year` into the query string. it will also put the following array into the session:

 `11mbmzx-columnSelectEnabled` => array:1 [▼
    0 => `year`
  ]

Query strings will be evaluated first to determined which columns should be displayed, after that the session, followed by the default.

closes #760

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,16 +4,14 @@ All notable changes to `laravel-livewire-tables` will be documented in this file
 
 ## [Unreleased]
 
-## [column-select-dev] - 2022-05-14
+## [column-select-dev] - 2022-05-16
 
 ### Added
 
 - Added functionality to bookmark or deep link column selection
-- Added functionality to identify different datatable components as unique in column selection and filter arrays
-
-### Changed
-
-- Simplified misc/multiple-tables.md in docs where additional steps are no longer needed. (Original steps still work and lead to same result)
+- Added functionality to identify different datatable components as unique in column selection
+- Added Funcitonality to configure query string alias
+- Added Funcitonality to configure session key for column selection (dataTableFingerprint)
 
 ## [2.7.0] - 2022-05-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to `laravel-livewire-tables` will be documented in this file
 
 ## [Unreleased]
 
+## [column-select-dev] - 2022-05-18
+
+### Added
+
+- Added functionality to select/desect all columns in column selection dropdown
+
 ## [column-select-dev] - 2022-05-16
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to `laravel-livewire-tables` will be documented in this file
 
 ## [Unreleased]
 
+## [dev-column-select] - 2022-05-14
+
+- Added functionality to bookmark or deep link column selection
+- Added functionality to identify different datatable components as unique in column selection and filter arrays
+
 ## [2.7.0] - 2022-05-07
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,6 @@ All notable changes to `laravel-livewire-tables` will be documented in this file
 
 ## [Unreleased]
 
-## [column-select-dev] - 2022-05-18
-
-### Added
-
-- Added functionality to select/desect all columns in column selection dropdown
-
 ## [column-select-dev] - 2022-05-16
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,16 @@ All notable changes to `laravel-livewire-tables` will be documented in this file
 
 ## [Unreleased]
 
-## [dev-column-select] - 2022-05-14
+## [column-select-dev] - 2022-05-14
+
+### Added
 
 - Added functionality to bookmark or deep link column selection
 - Added functionality to identify different datatable components as unique in column selection and filter arrays
+
+### Changed
+
+- Simplified misc/multiple-tables.md in docs where additional steps are no longer needed. (Original steps still work and lead to same result)
 
 ## [2.7.0] - 2022-05-07
 
--- a/docs/columns/column-selection.md
+++ b/docs/columns/column-selection.md
@@ -96,3 +96,18 @@ public function configure(): void
     $this->setRememberColumnSelectionDisabled();
 }
 ```
+
+### setDataTableFingerprint
+
+In order to idenfify each table and prevent conflicts on column selection, each table is given a unique fingerprint.
+This fingerprint is generated using the static::class name of the component. If you are reusing
+the same component in different parts of your application, you may need to set your own custom fingerprint.
+
+```php
+public function configure(): void
+{
+    // Default fingerprint is output of protected method dataTableFingerprint()
+    // Below will prepend the current route name
+    $this->setDataTableFingerprint(route()->getName() . '-' . $this->dataTableFingerprint());
+}
+```
--- a/docs/datatable/available-methods.md
+++ b/docs/datatable/available-methods.md
@@ -353,17 +353,6 @@ public function configure(): void
 }
 ```
 
-### setQueryStringAlias
-
-Set an alias for the query string.
-
-```php
-public function configure(): void
-{
-  $this->setQueryStringAlias('my-custom-alias');
-}
-```
-
 ## Relationships
 
 **Disabled by default**, enable to eager load relationships for all columns in the component.
--- a/docs/datatable/available-methods.md
+++ b/docs/datatable/available-methods.md
@@ -353,6 +353,17 @@ public function configure(): void
 }
 ```
 
+### setQueryStringAlias
+
+Set an alias for the query string.
+
+```php
+public function configure(): void
+{
+  $this->setQueryStringAlias('my-custom-alias');
+}
+```
+
 ## Relationships
 
 **Disabled by default**, enable to eager load relationships for all columns in the component.
--- a/docs/misc/multiple-tables.md
+++ b/docs/misc/multiple-tables.md
@@ -92,3 +92,14 @@ public function configure(): void
     $this->setQueryStringDisabled();
 }
 ```
+
+## Disabling column selection for multiple of the same component
+
+You should also disable the columns selection for those components so the column selection state does not get replaced by one or the other:
+
+```php
+public function configure(): void
+{
+    $this->setColumnSelectStatus(false);
+}
+```
--- a/docs/misc/multiple-tables.md
+++ b/docs/misc/multiple-tables.md
@@ -33,6 +33,55 @@ If you need the above, you should make them different components like so:
 <livewire:pending-users-table />
 ```
 
+## Introduction
+
+By default, your table has a name of `table`, as well as an internal array called `$table` which saves its state to the query string.
+
+The query string would look like this:
+
+```php
+// Under the hood
+public array $queryString = [
+    'table' => [
+        'search' => null,
+        'sort' => [],
+        ...
+    ],
+]
+```
+
+In order to have multiple tables on the same page, you need to tell it how to save the state of each table.
+
+## Setting the table name and data
+
+If you have multiple tables on the same page and you want them to have independent state saved in the query string, you must set a table name and data array.
+
+```php
+public string $tableName = 'users';
+public array $users = [];
+```
+
+The data array must be the same name as the table name. This data array will remain blank, I tried to create it dynamically in the query string but Livewire doesn't support that, so you have to define it yourself. It is a workaround until Livewire supports dynamic properties for the query string.
+
+Your query string will now look like this:
+
+```php
+// Under the hood
+public array $queryString = [
+    'users' => [
+        'search' => null,
+        'sort' => [],
+        ...
+    ],
+    // Other tables
+    'roles' => [
+        'search' => null,
+        'sort' => [],
+        ...
+    ],
+]
+```
+
 ## Disabling the query string for multiple of the same component
 
 If you must have multiple of the same component on the same page, you should disable the query string for those components so the query string state does not get replaced by one or the other:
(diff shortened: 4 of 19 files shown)


inmanturbo:
These changes are backwards compatible, however we should discuss the following:

- It may be worth considering to replace all the calls to `getTableName()` (for wire:ids) in the views with the new `dataTableFingerprint()`
- Users with more than one instance of the same component in the same app/site could still run into issues with the columns in the session, even on different pages, unless they implement their own `dataTableFingerprint()` method that maybe also appends the last segment of the route or URL or something like that.

inmanturbo:
These changes are backwards compatible, however we should discuss the following:

- It may be worth considering to replace all the calls to `getTableName()` (for wire:ids) in the views with the new `dataTableFingerprint()`
- Users with more than one instance of the same component in the same app/site could still run into issues with the columns in the session, even on different pages, unless they implement their own `dataTableFingerprint()` method that maybe also appends the last segment of the route or URL or something like that.
- some users could have stale links or sessions following the update (but that should clear up after the first visit).

rappasoft:
I haven't tested yet but it looks good. Just curious, why add to query string instead of boot:

1. Still alias getTableName in queryString but instead of userSelectedColumns change boot to:

```php
$this->{$this->tableName}` = [
    'sorts' => $this->{$this->tableName}['sorts'] ?? [],
    'filters' => $this->{$this->tableName}['filters'] ?? [],
    'columns' => $this->{$this->tableName}['userSelectedColumns'] ?? [],
];
```

Or something like that, unless I'm missing something, but it's cleaner to me that its under the same table key in the URL and not its own global thing?

inmanturbo:
I think it could be done that way yes. But with the fingerprint now the `$this->{$this->tableName}` syntax is really no longer required for multiple tables to work on the same page. So the user doesn't have to manually assign properties anymore. It could possibly be removed everywhere to simplify the code in the next version change and existing calls were only kept here as a sort of extra careful compatibility practice. I'm a little hesitant to add more calls to `$this->{$this->tableName}`. However it should work.

We also may have to check everywhere to make sure the array key `$this->{$this->tableName}['userSelectedColumns']`
is set to avoid errors. So for example I think the following?

```php
        // Set to either the default set or what is stored in the session
        $this->selectedColumns = count($this->userSelectedColumns) > 0 ?
            $this->userSelectedColumns :
            session()->get($this->getColumnSelectSessionKey(), $columns);
```
Becomes?

```php
        // Set to either the default set or what is stored in the session
        $this->selectedColumns = (isset($this->{$this->tableName}['userSelectedColumns']) && count($this->userSelectedColumns) > 0) ?
            $this->userSelectedColumns :
            session()->get($this->getColumnSelectSessionKey(), $columns);
```

inmanturbo:
Of course I could be wrong here. I'll have to run some tests using the `boot()` method outlined above.

inmanturbo:
Ah, I see. Works like this:
```php
        // Set to either the default set or what is stored in the session
        $this->selectedColumns = count($this->{$this->tableName}['columns']) > 0 ?
            $this->{$this->tableName}['columns'] :
            session()->get($this->getColumnSelectSessionKey(), $columns);
```
I'm going to push a change following your suggestion. It is a little better.

rappasoft:
I mean i'm all for removing that syntax if it's backward compatible.

inmanturbo:
@rappasoft 

> I haven't tested yet but it looks good. Just curious, why add to query string instead of boot:
> 
> 1. Still alias getTableName in queryString but instead of userSelectedColumns change boot to:
> 
> ```
> $this->{$this->tableName}` = [
>     'sorts' => $this->{$this->tableName}['sorts'] ?? [],
>     'filters' => $this->{$this->tableName}['filters'] ?? [],
>     'columns' => $this->{$this->tableName}['userSelectedColumns'] ?? [],
> ];
> ```
> 
> Or something like that, unless I'm missing something, but it's cleaner to me that its under the same table key in the URL and not its own global thing?

@rappasoft done and tested. (phpunit as well as user story with two tables on same page). Please have a look when you get a chance.

rappasoft:
Don't have too much time to test but I'm getting some funky behavior when I deselect a column:

`Cannot assign bool to property Rappasoft\LaravelLivewireTables\DataTableComponent::$selectedColumns of type array`

inmanturbo:
Ok thanks. I'll have a look when I get a chance.

inmanturbo:
Ok so looks like I can replicate the error using the demo. Looks like there is indeed some strange behavior, specifically when the columns aren't actually database columns. The check boxes don't have a value, and the value put into the session is `null`. I'm using `$column->getField()` in the checkbox. I supppose `getField()` must return null if there isn't a database field.

inmanturbo:
> Don't have too much time to test but I'm getting some funky behavior when I deselect a column:
> 
> `Cannot assign bool to property Rappasoft\LaravelLivewireTables\DataTableComponent::$selectedColumns of type array`

This was due to the use of `getField()` being used to identify selected columns in the checkboxes. This has been fixed by adding a `getSlug()` helper in `trait` `ColumnHelpers` , which makes a slug from the title, and using that. This results in the query string using slugged Titles, which is probably better anyway than showing the actual database column names on the frontend.

inmanturbo:
Here's how it must have worked before:

```bash
➜  laravel-livewire-tables-demo git:(master) ✗ php artisan tinker                   
Psy Shell v0.11.4 (PHP 8.1.5 — cli) by Justin Hileman
>>> md5(null)
PHP Deprecated:  md5(): Passing null to parameter #1 ($string) of type string is deprecated in laravel-livewire-tables-demoeval()'d code on line 1
=> "d41d8cd98f00b204e9800998ecf8427e"
>>> 
```
:thinking:

inmanturbo:
> I mean i'm all for removing that syntax if it's backward compatible.

I mean it should be doable. Here are the only potential issues I see for backwards compatibility:

Say some has two tables on one page, and followed the steps in the docs, setting `tableName` to "users", and adding a `users` property.

```php
public string $tableName = 'users';
public array $users = [];
```

The default of course is `table`. Now simply using `$this->table['filters']` for instance of course will ignore the `users` property. 

However I think is is actually a non-issue. There should be no errors because the `$table` property would still exist, and there's no way anyone could have removed it from the base component. So long as  the query strings are unique, and we makes sure the `wire:ids` will still be unique to each table. We could do that using the `dataTableFingerprint()` method in place of `getTableName()`.

The only thing left to deal with then is if someone has two of the _same_ table in a page: 

1. Already the docs instruct to disable the query string for these.
2. There's no mention in the docs but column selection won't work either.
3. There's no mention in the docs now but technically the wire:ids should be unique to each table.

All of these could be solved, and in fact `1`  should no longer be necessary (meaning the query strings would actually work), by Implementing and overriding the `dataTableFingerpint()` to make sure it's unique to each one of the instances somehow. 

However might that require an upgrade guide with new instructions for using multiple of the _same component_ on the same page? Which should probably only happen on a major version change. Idk. It might not matter since none of this stuff works right now anyway, and the default behavior would essentially be the same.

inmanturbo:
@rappasoft  Maybe this feature should be targeting the next major version anyway, since it's possible some people may have to update their links. If they are using the query string and have it hardcoded in anchor tags somewhere they could end up not with broken links or 404's or anything like that, but for instance filter unexpectedly not being set by the query string. 

What do you think? We could still just implement and merge the fingerprinting of the session key for now as a bug fix. Basically break this PR up into two parts.

rappasoft:
I mean we could let the current table names override the fingerprint, wouldn't that solve the link problem?

inmanturbo:
Wouldn't that be the same as just removing the alias in the query string?

inmanturbo:
So then I guess everything would be the same, except we would also have columns in the query string. And everyone could follow the existing steps in the docs for multiple tables on the same page. And of course the fingerprinting would take care of the session problem, for more than one table with matching columns.

inmanturbo:
Note we can't just use the table name in the session. Because the default table names are always the same, so going from one table to another on any page you have conflicts.

inmanturbo:
You know, another option is we could use the table name as an alias and implement something like `getQueryStringAlias` that return the table name by default but would also it could be overridden. Perhaps even with something like

```php
{
    $this->setQueryStringAlias('users-table2');
}
```

inmanturbo:
I'm going to mark this as a draft work on it throughout the week in my spare time. @rappasoft just to go over where I think we're at with it now. We will default to the table name in the query string, with a configurable alias. and allow for the fingerprint to be set/overridden using `configure()` as well, and add to the docs to reflect this new functionality, with an api like so.

```php
{
    $this->setQueryStringAlias('users-table2'); // defaults to $tableName
    $this->setDataTableFingerprint('my-custom-table-fingerpint'); // defaults to hash made from the base 36 of the serialized crc32 of static::class
}
```

Everything works as expected so right now, so far as I can tell. The issue you ran into when testing has been fixed. But I'm thinking think we should probably preserve everyone's links for now? And it would be cool to set up this feature with an api that can be easily documented and understood by the community.

inmanturbo:
@rappasoft due to the extensibility of working with the `configure()` method, this actually took less time than I expected! Please look it over when you can.

inmanturbo:
@Ceepster14 I think we can consider this branch pretty stable at this point. Is there any chance you could test this in your project and confirm that it close https://github.com/rappasoft/laravel-livewire-tables/issues/760 ?

```bash
composer require rappasoft/laravel-livewire-tables:column-select-dev
```

Ceepster14:
Hi guys,

Great work! I've given this fix a bit of a thrashing and it all seems to hold up. No more missing columns. Thanks for all your efforts!

Chris

inmanturbo:
unfortunately one part of this feature still isn't working properly:

```php
{
    $this->setQueryStringAlias('users-table2'); // defaults to $tableName
}
```

This indeed sets an alias, however on page refresh or a fresh visit it clears the query string. I thought that maybe it was because I hadn't added an actual public property to the `DataTableComponent` for $queryStringAlias. But I've tried adding it and it doesn't fix the problem.

Here is how the alias is set during configuration:
```php
    public function setQueryStringAlias(string $queryStringAlias): self
    {
        $this->queryStringAlias = $queryStringAlias;

        return $this;
    }
```

And here is how it is used in the query string:

```php
    public function queryString(): array
    {
        if ($this->queryStringIsEnabled()) {
            return [
                $this->getTableName() => ['except' => null, 'as' => $this->getQueryStringAlias()],
            ];
        }

        return [];
    }
```
And here is the `getQueryStringAlias method
```php
    public function getQueryStringAlias(): string
    {
        return $this->queryStringAlias ?? $this->getTableName();
    }
```

So far I've tried adding `public $queryStringAlias` to the component:
```diff

    public array $table = [];
    public $theme = null;
+   public $queryStringAlias = null;
```

But this doesn't seem work. Any ideas or help would be appreciated. Marking this as a draft for now. WIP

inmanturbo:
Hmm ... Looks like it's back to the same problem as here https://github.com/livewire/livewire/discussions/4300.

hard coding the alias works fine:

```
public $queryStringAlias = 'alias';
```

but trying to set it at runtime doesn't seem to work.

inmanturbo:
This seems to have done the trick. Any reason we can't do this? 
```diff
    /**
     * Runs on every request, immediately after the component is instantiated, but before any other lifecycle methods are called
     */
    public function boot(): void
    {
        $this->{$this->tableName} = [
            'sorts' => $this->{$this->tableName}['sorts'] ?? [],
            'filters' => $this->{$this->tableName}['filters'] ?? [],
            'columns' => $this->{$this->tableName}['columns'] ?? [],
        ];
        
        // Set the filter defaults based on the filter type
        $this->setFilterDefaults();
+      $this->configure();
    }

    /**
     * Runs on every request, after the component is mounted or hydrated, but before any update methods are called
     */
    public function booted(): void
    {
-      $this->configure();
        $this->setTheme();
        $this->setBuilder($this->builder());
        $this->setColumns();

        // Make sure a primary key is set
        if (! $this->hasPrimaryKey()) {
            throw new DataTableConfigurationException('You must set a primary key using setPrimaryKey in the configure method.');
        }
    }
```

inmanturbo:
> Does this PR fix this issue #760?

Yes it does. And it should be good to go. I just wanted Anthony to look over it before merging it in. I'm sure he'll review the open PR's before the next release, and he releases often. He just got sick and things got a little backed up for him. Should be back soon.

rubensrocha:
I look forward to the release of a new version to include this very important fix. :)

PR #781, opened 2022-05-17
Column::make: return type static

levu42:
`LinkColumn::make()->title()` gives currently an error in PHPStan in my project because it expects `LinkColumn::make` to return a `Column` even though it returns a `LinkColumn`.

I saw in your composer.json file that you also maintain compatability with 7.4, so I only changed the return type in the doc block. Ideal would of course be to also have `static` as return type also in the code itself.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/src/Views/Column.php
+++ b/src/Views/Column.php
@@ -84,7 +84,7 @@ class Column
      * @param  string  $title
      * @param  string|null  $from
      *
-     * @return Column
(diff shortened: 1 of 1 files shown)


CristhoferMF:
this will be great. to prevent lint error messages like:
undefined method 'functionxyz'

PR #783, opened 2022-05-19
Select deselect all columns

inmanturbo:
I've notice while working with the column select feature it can get a little tedious having to uncheck and check each column every time.
- This adds functionality to check all and uncheck all columns. 
- Checking all columns just sets it back to the default, so it also clears the "columns" query string (but not the filters, etc).
- Checking all will not select columns that are configured as unselected. Those will have to be checked, or selected, manually, as is the default.
- This adds the feature for all themes

### DIFF
--- a/docs/datatable/available-methods.md
+++ b/docs/datatable/available-methods.md
@@ -353,17 +353,6 @@ public function configure(): void
 }
 ```
 
-### setQueryStringAlias
-
-Set an alias for the query string.
-
-```php
-public function configure(): void
-{
-  $this->setQueryStringAlias('my-custom-alias');
-}
-```
-
 ## Relationships
 
 **Disabled by default**, enable to eager load relationships for all columns in the component.
--- a/src/DataTableComponent.php
+++ b/src/DataTableComponent.php
@@ -71,7 +71,6 @@ abstract class DataTableComponent extends Component
         
         // Set the filter defaults based on the filter type
         $this->setFilterDefaults();
-        $this->configure();
     }
 
     /**
@@ -79,6 +78,7 @@ abstract class DataTableComponent extends Component
      */
     public function booted(): void
     {
+        $this->configure();
         $this->setTheme();
         $this->setBuilder($this->builder());
         $this->setColumns();


inmanturbo:
Peek 2022-05-20 10-44

inmanturbo:
@rappasoft can you please take a look and see what you think about the implementation here when you get a chance? The plan is to merge this into `column-select` before merging `column-select` into `develop`. Alternatively, If we could squash merge `column-select` into `develop` first, then merge this into `develop` afterwards so the commit history make more sense.

The GIF above demonstrates the UX.

inmanturbo:
note checking all doesn't check `My Link`, this is because `My Link` is configured as `deselected` by default.
```php
            LinkColumn::make('My Link')
                ->title(fn($row) => 'Edit')
                ->location(fn($row) => 'https://'.$row->id.'google.com')
                ->attributes(function($row) {
                    return [
                        'class' => 'underline text-blue-500',
                    ];
                })
                ->deselected(),
```
This allows checking all to simply set the defaults, which means it clears the query string as well as the session.

```php     public function updatedSelectedColumns(): void
    {
        // The query string isn't needed if it's the same as the default
        if ($this->allDefaultVisibleColumnsAreSelected() && $this->allSelectedColumnsAreVisibleByDefault()) {
            $this->selectAllColumns();
        } else {
            $this->{$this->tableName}['columns'] = $this->selectedColumns;
            session([$this->getColumnSelectSessionKey() => $this->{$this->tableName}['columns']]);
        }
    }
```

The case where we have columns that available to be checked but aren't actually visible by default is the only case that requires a little consideration. If none of the columns are configured as `deselected` by default, It's just your typical Check/Uncheck all UX.

inmanturbo:
Peek 2022-05-20 11-15

inmanturbo:
This is sort of a minimalist approach to the design here. Another option would be to have a checkbox that always checks all and puts all of them into the query string and session, and a separate button somewhere which sets the defaults, perhaps even one at top level in the tool bar to restore defaults for all options if any are dirty (filters, search, column select, page).

saleemepoch:
Hi @inmanturbo, thanks for the contribution.

I was wondering when this will be merged to master? A client of ours did express that the expected behaviour of Select All, should be select all and shouldn't exclude deselected columns.

Many thanks

lrljoe:
It's showing as closed, so it won't be merged if the feature wasn't previously, it would need reopening, a review, and testing.

inmanturbo:
> Hi @inmanturbo, thanks for the contribution.
> 
> I was wondering when this will be merged to master? A client of ours did express that the expected behaviour of Select All, should be select all and shouldn't exclude deselected columns.
> 
> Many thanks

This feature has been merged. This PR was closed in favor of another. See https://github.com/rappasoft/laravel-livewire-tables/pull/802

inmanturbo:
Here is the commit at the time it was merged into master:
https://github.com/rappasoft/laravel-livewire-tables/commit/4ca262d15bfebebbd80f2620b239d208a728ae06

PR #802, opened 2022-06-08
Select deselect all columns

inmanturbo:
I've notice while working with the column select feature it can get a little tedious having to uncheck and check each column every time.
- This adds functionality to check all and uncheck all columns. 
- Checking all columns just sets it back to the default, so it also clears the "columns" query string (but not the filters, etc).
- Checking all will not select columns that are configured as unselected. Those will have to be checked, or selected, manually, as is the default.
- This adds the feature for all themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to `laravel-livewire-tables` will be documented in this file
 
 ## [Unreleased]
 
+## [column-select-dev] - 2022-05-18
+
+### Added
+
+- Added functionality to select/desect all columns in column selection dropdown
+
 ## [column-select-dev] - 2022-05-16
 
 ### Added
--- a/docs/datatable/available-methods.md
+++ b/docs/datatable/available-methods.md
@@ -353,17 +353,6 @@ public function configure(): void
 }
 ```
 
-### setQueryStringAlias
-
-Set an alias for the query string.
-
-```php
-public function configure(): void
-{
-  $this->setQueryStringAlias('my-custom-alias');
-}
-```
-
 ## Relationships
 
 **Disabled by default**, enable to eager load relationships for all columns in the component.
--- a/resources/views/components/tools/toolbar.blade.php
+++ b/resources/views/components/tools/toolbar.blade.php
@@ -246,6 +246,26 @@
                         >
                             <div class="bg-white rounded-md shadow-xs dark:bg-gray-700 dark:text-white">
                                 <div class="p-2" role="menu" aria-orientation="vertical" aria-labelledby="column-select-menu">
+                                    <div>
+                                        <label
+                                            wire:loading.attr="disabled"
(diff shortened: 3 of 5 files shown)


inmanturbo:
Peek 2022-05-20 10-44

inmanturbo:
@rappasoft can you please take a look and see what you think about the implementation here when you get a chance? The plan is to merge this into `column-select` before merging `column-select` into `develop`. Alternatively, If we could squash merge `column-select` into `develop` first, then merge this into `develop` afterwards so the commit history makes more sense.

The GIF above demonstrates the UX. (including the query string)

inmanturbo:
note checking all doesn't check `My Link`, this is because `My Link` is configured as `deselected` by default.
```php
            LinkColumn::make('My Link')
                ->title(fn($row) => 'Edit')
                ->location(fn($row) => 'https://'.$row->id.'google.com')
                ->attributes(function($row) {
                    return [
                        'class' => 'underline text-blue-500',
                    ];
                })
                ->deselected(),
```
This allows checking all to simply set the defaults, which means it clears the query string as well as the session.

```php 
    public function selectAllColumns()
    {
        $this->{$this->tableName}['columns'] = []; 
        $this->forgetColumnSelectSession();
    }
```

The case where we have columns that are available to be checked but aren't actually visible by default is the only case that requires a little extra consideration. If none of the columns are configured as `deselected` by default, It's just your typical Check/Uncheck all UX.

inmanturbo:
This is sort of a minimalist approach to the design here. Another option would be to have a checkbox that always checks all and puts all of them into the query string and session, and a separate button somewhere which sets the defaults, perhaps even one at top level in the tool bar to restore defaults for all options if any are dirty (filters, search, column select, page).

saleemepoch:
@inmanturbo, thanks for referring me to this thread.

What can I do to ensure that Select All selects all columns, including the deselected ones?

inmanturbo:
> ```
> selectAllColumns()
> ```

You can override `selectAllColumns()` and inject our logic there. This should do the trick:

```php
    public function selectAllColumns()
    {
        $this->{$this->tableName}['columns'] = collect($this->getColumns())
        ->filter(function ($column) {
            return $column->isVisible() && $column->isSelectable();
        })
        ->map(fn ($column) => $column->getSlug())
        ->values()
        ->toArray();

        session([$this->getColumnSelectSessionKey() => $this->{$this->tableName}['columns']]);
    }
```
Note, this will create a long query string. Due to how the component uses query strings to allow deep linking or bookmarking a table view with selected columns displayed, and the fact that deselected columns need to be explicitly marked as selected in order to be displayed, it is not possible to have `selectAllColumns()` also select columns  which are not selected by default without cluttering up the query string. This is why it was decided to have the default way simply set the defaults.

saleemepoch:
@inmanturbo thanks very much for your help.

It seems to work and thanks for your note on the long query string. So far in our case I haven't noticed any performance issue so this will do.

PR #844, opened 2022-07-14
Add support for MorphOne relationships

DanielGSoftware:
**Take the following relationship**
Model `Address`, which has a `morphTo()` method and has `addressable_id` and `addressable_type` columns.
Model `User` , which has `MorphOne()` that points to the address model.

If I have a user table, and I want to display the street of the address, I would do `Column::make('street', 'address.street')`.

This will give the following exception: `Rappasoft\LaravelLivewireTables\Views\Column::setTable(): Argument #1 ($table) must be of type string, null given, called in /Users/danielgebben/Documents/LenderSpender/platform/vendor/rappasoft/laravel-livewire-tables/src/Traits/Helpers/ColumnHelpers.php on line 24`.

This is because MorphOne is not supported when making the columns.

### All Submissions:

### New Feature Submissions:

4. Did you test locally to make sure your feature works as intended? Yes

### Changes to Core Features:

Not yet, I'd first like this to pass and then I'll add tests


the-turk:
thank you so much for finding this, you saved my day

Francesco-Cirillo:
thank you for the change.
I need it in my web project, but I don't want to edit my vendor directory: I have to wait he pull request will be accepted?

KSneijders:
Just manually applied the small changes to test and it works flawlessly! Thanks @DanielGSoftware! ❤️ 

@rappasoft What is necessary for this to be merged?

lrljoe:
Tested against core, all appears working @rappasoft  - this seems okay to commit

PR #883, opened 2022-08-19
Add check for table 'from' field inside ImageColumn

lvismer:
Let's say you have an **image** field on your table. The current solution for **ImageColumn** will not query the **image** column from the database because somehow the existing label callback returns null which in turn results in *$this->field* being null.


rappasoft:
This change does not make it work though. I'd imagine the image would just automatically appear?

PR #911, opened 2022-09-14
Filter excluded columns from select

howdu:
I found that `excludeFromColumnSelect` wasn't working corectly. 

E.g 
```php
Column::make('Actions', 'actions')
                ->excludeFromColumnSelect()
                ->view('tables.columns.actions')
```


howdu:
Causes issues when filtering columns.

cswni:
This is a usefull improvements. Really works!

Issue #874, opened 2022-08-11
Problems with loading relationships


Oleksandr-Moik:
Case, I have to models - `User` and `Role`. `User` **belongsToMany** `Roles`. 

And I want to disblay all user roles with this column definition:
```php
Column::make('User roles', 'roles.name')
    ->eagerLoadRelations()
    ->format(function ($roles) {
        return $roles->implode(', ');
    })
```

Buy I catch an exception:
![Exception screenshot](https://user-images.githubusercontent.com/50796878/184252125-cd0fb08a-f1a1-4b5c-af2f-e0533c29521e.png)

During debugging, it turned out that the problem is in determining table name in method [`WithData::getTableForColumn()`](https://github.com/rappasoft/laravel-livewire-tables/blob/master/src/Traits/WithData.php#L134-L136) 

This issue is related pull about **Support MorphOne**: #844 

Also, I think we should add support other relationships types, not only **HasOne** or **BelongsTo**

fbrettnich:
I have a similar problem. Would also be happy about `hasMany` support.

Oleksandr-Moik:
If anyone has a similar problem and is looking for a solution, here is a small fix or temporary solution _(based on my case)_.

Define column like this:

```php
Column::make('User roles', 'roles')
    ->label(function (User $row): string {
        return $row->roles->pluck('name')->implode(', ');
    })
```

And add eager `roles` loading to query builder, to prevent lazy loading for each user row

```php
public function builder(): Builder
{
    return User::query()->with('roles');
}

```

tonypartridge:
Yeah it's a very poor performance way to handle it, you are adding a query for every single row. 

If you add it into the Builder method of the component it uses two additional queries. But for some reason that I am unable to figure out, we cannot select the 'AS' value from the with on the query builder because the column builder always references the Parent models table first and 'AS' values will not work.

vaskomichal:
I have the same problem. I selected column with custom join used 'AS' alias and now is not possible to show this alias column in table.

tonypartridge:
Its such a shame, as it is a massive performance increase and something that should be easily catered for, simply not applying table_name before each field would suffix like:

->applyTableNamePrefix(false)

PR #957, opened 2022-10-24
The query-select for the column can be modified or omitted.

LWlook:
### All Submissions:

### New Feature Submissions:

This pull requests let you hide the table-name prefix in front of some selects in the query. This is helpful to enable sorting and filtering for additional added columns (like with a join). With this pull it is also possible to mitigate some problems related to relationships. (Requested in Issue #874 )

If this pull request would be merged, than I can add the documentation too.

### DIFF
--- a/src/Traits/WithData.php
+++ b/src/Traits/WithData.php
@@ -117,8 +117,11 @@ trait WithData
         }
 
         foreach ($this->getSelectableColumns() as $column) {
-            $tablePrefix = ($column->getApplyTableNamePrefix() ? $column->getColumn() . ' as ' : '');
-            $this->setBuilder($this->getBuilder()->addSelect($tablePrefix . $column->getColumnSelectName()));
+            if ($column->getModifiedQuerySelect() === "") continue;
+
+            $selectQuery = ($column->getModifiedQuerySelect() === null ?
+                $column->getColumn() . ' as ' . $column->getColumnSelectName() : $column->getModifiedQuerySelect());
+            $this->setBuilder($this->getBuilder()->addSelect($selectQuery));
         }
 
         return $this->getBuilder();
--- a/src/Traits/WithData.php
+++ b/src/Traits/WithData.php
@@ -13,7 +13,7 @@ trait WithData
     public function getRows()
     {
         $this->baseQuery();
-        
+
         return $this->executeQuery();
     }
 
@@ -117,7 +117,8 @@ trait WithData
         }
 
         foreach ($this->getSelectableColumns() as $column) {
-            $this->setBuilder($this->getBuilder()->addSelect($column->getColumn() . ' as ' .$column->getColumnSelectName()));
+            $tablePrefix = ($column->getApplyTableNamePrefix() ? $column->getColumn() . ' as ' : '');
+            $this->setBuilder($this->getBuilder()->addSelect($tablePrefix . $column->getColumnSelectName()));
         }
 
         return $this->getBuilder();
--- a/src/Views/Column.php
+++ b/src/Views/Column.php
@@ -55,7 +55,7 @@ class Column
     protected bool $footer = false;
     protected $footerCallback;
     protected bool $clickable = true;
-    protected bool $applyTableNamePrefix = true;
+    protected ?string $modifiedQuerySelect = null;
 
     /**
      * @param  string  $title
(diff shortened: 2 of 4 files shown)


tonypartridge:
Amazing stuff! I was just going to dig down on this now. The problem with how V2 works is that because we have already defined the select in the query builder it's trying to select it again as part of column make.

My example is:

`php   ::query()->withSum('CYTD as cytd_total', 'total_amount');`
                    
cytd_total has been set, but we then try setting it as cytd with:

 ```php
           Column::make('Total CYTD', 'cytd_total')
                ->format(fn($v) => number_format((float) $v, 2))
                ->sortable(),
```
                
We could look at a different setter? 

isColumnPreSelected

?

**Comment by tonypartridge** [comment]
Maybe addQuerySelect(false) seems the most readable, then you wouldn't need to include table prefix option?

**Comment by LWlook** [comment]
Yes, addQuerySelect(false) would probably be more readable. Or alternativ `excludeFromQuerySelect()`.
But in my changes I only exclude the prefix, not the whole column name. Should work in your example too.
Would than look like that:

```php
          Column::make('Total CYTD', 'cytd_total')
               ->addQuerySelect(false)
               ->format(fn($v) => number_format((float) $v, 2))
               ->sortable(),
```

@rappasoft What is your opinion on this pull request?

tonypartridge:
So you example doesn't work for me because when you include it in the queryBuilder as a join / sum then you define it and it trys to get it again with the same prefix as it's already been defined. Your code isn't checking if it has been defined which is where the excludeFromQuerySelect() or addQuerySelect(false) works. I've overrode the code function for now to work on it.

But @rappasoft  seems to have been silently recently :'(

LWlook:
@tonypartridge What do you think about having a method like `modifyQuerySelect(String? select)`? So you can remove the select completely with `modifyQuerySelect()` or `modifyQuerySelect("")` and I don't have to write every join column to the additionalSelects so I can write them directly to the column definition. (Has the benefit that everything is in one place, and the select will not be triggered if the column is hidden).
Additionally with this method someone has still the ability to remove just the prefix.

TonyPartridgeR:
That sounds much more viable long term, we wouldn't need to do the prefix changing as we are selecting to column. we can empty it out. 

Spoke with Rappasoft too, he's just a bit busy at the moment but is hoping to spend some time on this project soon. 

If you are happy to do a PR we can review and test ;-)

LWlook:
@tonypartridge What do you say to my changes?

TonyPartridgeR:
They look good, but I need to test them out today. I’ve just been a bit waylaid!

tonypartridge:
Looking at this, it would make more sense if we could use the query builder. IT only allows string based SQL

tonypartridge:
Example:

  (
    select 
      sum(
        `laravel_reserved_0`.`total`
      ) 
    from 
      `records` as `laravel_reserved_0` 
    where 
      `records`.`pp_id` = `laravel_reserved_0`.`pp_id` 
      and `period` between '2022-01-01' 
      and '2022-11-03' 
      and `laravel_reserved_0`.`deleted_at` is null
  ) as `cytd_calc`
  
  Cannot be included as far as I ca see within the modifyQuerySelect method. I can now reference the select if I am handling it in the builder function. But the builder doesn't get the filtered values. So ideally we need support to allow optimising the select. In my case I want to pass a filter into the query select essentially, which is very close but not yet workable.

TonyPartridgeR:
Thanks @LWlook, I've moved to filamentphp tables and forms now since it actually does all this.

LWlook:
Probably I should also take a look at another library.

TonyPartridgeR:
It is a shame as it has great potential! But once you checkout filamentphp it's insanely good. Hit me up if you need any pointers.

rappasoft:
Sorry i've been MIA. I'm trying to get into this PR but I think I need a crash course on what we're trying to do.

tonypartridge:
@rappasoft the idea is to allow modifying the select Query of the column. Example is that if you build the select or join into the builder but then define the column, it can be needed to modify the select. The current way isn't possible so this method was proposed. 

I've moved to Filamentphp Forms and Tables now which allowed me to do it and with their admin panel outside the need for forms/tables it just makes everything rapid.

PR #1021, opened 2023-01-13
Default value for select when options() are driven from an query/collection.

lrljoe:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

The basic premise of this is for when users are generating the content of filters using builder/query.  The limitation of this is that there is no "all" function.  This feature allows them to set a "top-of-list" item (i.e. SELECT ALL), which will set to "all" and remove the filter from being applied.

### DIFF
--- a/docs/filters/creating-filters.md
+++ b/docs/filters/creating-filters.md
@@ -38,6 +38,27 @@ public function filters(): array
 
 You should supply the first option as the default value. I.e. nothing selected, so the filter is not applied. This value should be an empty string. When this value is selected, the filter will be removed from the query and the query string.
 
+When creating options from a Query or Builder function, you can alternatively use the setFirstOption() below to set this first value instead of having to merge arrays.
+
+```php
+use Rappasoft\LaravelLivewireTables\Views\Filters\SelectFilter;
+
+public function filters(): array
+{
+    return [
+        SelectFilter::make('Tags')
+            ->options(
+                Tag::query()
+                    ->orderBy('name')
+                    ->get()
+                    ->keyBy('id')
+                    ->map(fn($tag) => $tag->name)
+                    ->toArray()
+            )
+            ->setFirstOption('All Tags'),
+    ];
+}
+```
 ### Option Groups
 
 To use `<optgroup>` elements, pass a nested array of options to the select filter.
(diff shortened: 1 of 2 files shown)


rappasoft:
Is this finished? It doesn't do anything. Plus, there is already a default all checkbox with multiselect filters.

lrljoe:
I may have been massively confused when I wrote this.  But looking at my scrawled notes, the logic behind this is when generating the list of options from a query/builder, rather than directly specifying them.  

Rather than having to do an array_merge on the generated array of options to add a default/first option, this can be used to set that default/first option in the list.

I'll have another look at it and see if it actually does what it says on the tin, I may have selected the wrong branch.

ccsliinc:
@lrljoe this code does not do anything.

ccsliinc:
@lrljoe I think you need to add something like this to the code.  Also why don't you add the same logic to the select filter.

`
    public function options(array $options = []): MultiSelectDropdownFilter
     {
         $this->options = $this->firstOption ? ['' => $this->firstOption] + $options : $options;

         return $this;
     }
`

lrljoe:
> @lrljoe I think you need to add something like this to the code.  Also why don't you add the same logic to the select filter.
> 
> ```php
>     public function options(array $options = []): MultiSelectDropdownFilter
>      {
>          $this->options = $this->firstOption ? ['' => $this->firstOption] + $options : $options;
> 
>          return $this;
>      }
> ```

Yep, I messed up my branches, will merge and commit this weekend so that this pull updates

lrljoe:
Reminder to self to finish fixing this.

rappasoft:
Is this not the same as adding `'' => 'All'` as the first option anyway?

lrljoe:
It was more for when you generate the array outside of the filters() method.

For example, I setup a fairly complex array on first load, either from cache or query and store it in a public variable, and then use that variable in the options() method of a filter, rather than shoving a ton of code into that method

I'm happy enough with this not being pulled as anyone doing that should know enough to be able to add a first item as select all anyway.

rappasoft:
I usually do `array_merge(['' => 'All'], $otherComplexArray)` but i have no problem merging as a helper utility if it works.

lrljoe:
Hang fire on it, I'll clean it up a little in terms of docs to make it obvious what the use case is and can stick it in next time

rappasoft:
Feel free to finish it, i'm good with there being more than one way to do things.

lrljoe:
Cool will take a look later this week and do any tidying/tweaks needed.

PR #1023, opened 2023-01-15
Adding setAllVisibleSelected functionality  (Pending Fixes)

lrljoe:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

**Additional Functionality**
This adds a new option to **add** the items on the current page to the "selected" array, which gives options to users working with larger quantities of results.   This would allowing for someone to add the first X pages of results, or select all items on a particular page for use with a Bulk Action.

Worth noting that this one **"ADDS"**  to the existing selection, rather than resetting and rebasing.  I was in half-minds whether to have this reset after a page navigation, or retain the existing selected, having been on both sides of the result.

UX Displays as follows - all languages now have translations for this
image

Choosing "All On This Page" **ADDS** to your current Selected with those visible on the page (that aren't already selected).


rappasoft:
This mostly works, however, when you select all on a page, then change pages to select more, all previous checkboxes are still selected.

rappasoft:
I might might not have explained well:

- Check 1 box on page 1
- Click select all on page
- Go to page 2
- All checkboxes are still selected

lrljoe:
Ah yep, I understand the issue now, weirdly didn't experience this when I was testing, so I think I've messed up a merge somewhere!
Will fix when I get a chance over next day or two and update here.

lrljoe:
Should be fixed now, but well worth double checking as I need to clean up my repo for this package.

I plan to convert Bulk Actions into a proper AlpineJS piece to remove the constant calls when you select/deselect an item, unless you object?

rappasoft:
Definitely still not working correctly, for example:

- Select 1 on page 1, then hit select all for this page: 11 selected but only 10 checked
- Select page 1, then go to page 9 and select all on that page: 20 selected but nothing checked

This was a hard problem which is why I left it as only select all if I remember correctly. I appreciate you trying to figure it out though.

lrljoe:
Yeah I think I've stuffed up with my branches here and gotten myself confused about which one I'm committing to.

In my test environment for that use case with a 10-per-page pagination setting I'm getting:
Select 1 on page 1, Hit "select all for this page", results in 10 selected on the page, with 10 boxes checked, and IDs match against $selected in the backend.

Next test case was select one on page 1, then select all on page 1, go to page 9, select one, select all on this page, 20 selected, 20 checked, IDs matching, uncheck one on page 9, then uncheck one on page 1, results in 18 (as expected).

Let me know if the current commit actually looks sane, otherwise I'll scrap the branch and port the code into a fresh one.

rappasoft:
Here's a video of how it works for me, as you can see the counts are off and the checkboxes aren't populating right:


https://user-images.githubusercontent.com/6026504/213074827-161b498c-6576-425b-9dcf-f368b19fa3d6.mov


Don't kill yourself on this one, it would be nice, but it's not high on my priority list.

lrljoe:
Just checking that you have grabbed the latest commit for your test.  If so I'll take another look at it later in the week and figure out what's going on!

lrljoe:
Have a new version of this, which is mostly AlpineJS with a minor tweak, just looking at how I can clean it up a little further

lrljoe:
Check out demo on [https://filterdemo.lowerrocklabs.com/](https://filterdemo.lowerrocklabs.com/) - believe this should be working smoothly now!

lrljoe:
Replaced with #1112

PR #1036, opened 2023-02-01
Custom view prepend

SimonMacIntyre:
The currently existing `customView` is only suitable for modals and things at the bottom of the page. I really needed something like this as well so that I can do what I have shown in the video comment I added below. 

I would propose renaming `customView` to `customViewAppend` but not worth breaking changes obviously. 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/misc/custom-markup.md
+++ b/docs/misc/custom-markup.md
@@ -13,3 +13,14 @@ public function customView(): string
     return 'includes.custom';
 }
 ```
+
+If you would like to append any custom markup right before the start of the component, you may use the `customViewPrepend` method and return a view.
+
+This is good for loading extra content such as dismissible alerts or flash notifications.
+
+```php
+public function customViewPrepend(): string
+{
+    return 'includes.custom';
+}
+```
--- a/resources/views/datatable.blade.php
+++ b/resources/views/datatable.blade.php
@@ -1,4 +1,7 @@
 <x-livewire-tables::wrapper :component="$this">
+    @isset($customViewPrepend)
+        @include($customViewPrepend)
+    @endisset
     <x-livewire-tables::tools>
         <x-livewire-tables::tools.sorting-pills />
         <x-livewire-tables::tools.filter-pills />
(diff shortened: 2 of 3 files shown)


SimonMacIntyre:
Example of the feature this unlocks:

https://user-images.githubusercontent.com/1260938/216085328-db764dba-69ad-412a-8e02-424840c35b95.mov

SimonMacIntyre:
I was not aware of `before-toolbar`, and now I am unsure if this is different at all. This was just after I was looking at `Custom View` in isolation, and wondering why an `after` block existed but not a `before`, so I made this. So I am not sure if `before-toolbar` is the exact same functionally as this, or if this offers anything new at all.

lrljoe:
As mentioned on the [Discussion](https://github.com/rappasoft/laravel-livewire-tables/discussions/1035), I'd recommend adding an additional configurableArea if before-toolbar doesn't suit (which it may not due to the filter/sorting pils), over this.

SimonMacIntyre:
> I was not aware of `before-toolbar`, and now I am unsure if this is different at all. This was just after I was looking at `Custom View` in isolation, and wondering why an `after` block existed but not a `before`, so I made this. So I am not sure if `before-toolbar` is the exact same functionally as this, or if this offers anything new at all.

After testing this, it seems there is a difference in how the loading/reaction to variable changes between this and `before-toolbar`. 

Using my already working example, I simply commented out my `customViewPrepend`, and instead used 
```
$this->setConfigurableAreas([
            'before-toolbar' => 'data-table-error'
        ]);
```

But what was interesting, is it did not seem to reload or pick up the variable change and thus show, so it never shows. 
I'm sure there is some way to control that, but to me this means this is different in some way. And I prefer the simplicity of my PR, if I'm being honest, which works as expected intuitively with changing variables. 

I wonder if it is because my custom view prepend is a simple blade view, rather than a livewire component?

lrljoe:
Try using $this->variableName instead, and I imagine it'll work smoothly.

SimonMacIntyre:
> Try using $this->variableName instead, and I imagine it'll work smoothly. Or entangle it with alpine and do it that way.
> 
> I built something very similar for displaying the download links for Exports, but using the before-toolbar area.

Aha.... yes that did it. 

Thanks a lot @lrljoe, appreciate you always being here to help!

lrljoe:
No problem, I'll update those docs in my next PR, and might add a pointer in there around variable access, as no doubt others have had the same issue!

PR #1037, opened 2023-02-01
New configurable area 'before-tools', to be able to place above also the pills

SimonMacIntyre:
Adds a configurable `before-tools` section, to be able to be at the very tip-top even above filter/sort pills. 

Example video attached. 

https://user-images.githubusercontent.com/1260938/216110230-7dc01648-95a3-4c75-a35c-6b82a3618331.mov

Related to a previous PR attempting the same thing in a different way: https://github.com/rappasoft/laravel-livewire-tables/pull/1036

### All Submissions:

* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests and did you add any new tests needed for your feature?
2. [ ] Did you update all templates (if applicable)?
3. [x] Did you add the [relevant documentation](https://github.com/rappasoft/laravel-livewire-tables-docs) (if applicable)?
4. [x] Did you test locally to make sure your feature works as intended?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1060, opened 2023-02-21
Implemented ability to hide toolbar and entire data table when empty

gRoberts84:
For purposes of DRY, I am using the same data table component in multiple areas of the site, some of which don't need the toolbar etc (disabling individual components hides them but still keeps the wrappers) and some don't need to appear at all if there are no results to show. 

In addition to this, I also added the ability to include a custom view before the data table, so you can add things like titles, that if combined with hiding when empty, shouldn't show either.

Happy to make as many changes as needed to make it acceptable.

---

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/resources/views/components/tools.blade.php
+++ b/resources/views/components/tools.blade.php
@@ -5,11 +5,11 @@
 @endphp
 
 @if ($theme === 'tailwind')
-    <div class="flex-col">
+    <div class="flex-col tools">
         {{ $slot }}
     </div>
 @elseif ($theme === 'bootstrap-4' || $theme === 'bootstrap-5')
-    <div class="d-flex flex-column">
+    <div class="d-flex flex-column tools">
         {{ $slot }}
     </div>
 @endif
--- a/resources/views/datatable.blade.php
+++ b/resources/views/datatable.blade.php
@@ -1,13 +1,19 @@
+@if (!$this->getHideOnEmptyStatus() || ($this->getHideOnEmptyStatus() && !empty($rows)))
+@isset($beforeComponentView)
+    @include($beforeComponentView)
+@endisset
 <x-livewire-tables::wrapper :component="$this">
     @if ($this->hasConfigurableAreaFor('before-tools'))
         @include($this->getConfigurableAreaFor('before-tools'), $this->getParametersForConfigurableArea('before-tools'))
     @endif
 
+    @if ($this->getToolbarStatusIsEnabled())
     <x-livewire-tables::tools>
         <x-livewire-tables::tools.sorting-pills />
         <x-livewire-tables::tools.filter-pills />
         <x-livewire-tables::tools.toolbar />
     </x-livewire-tables::tools>
+    @endif
 
     <x-livewire-tables::table>
         <x-slot name="thead">
@@ -69,3 +75,4 @@
         @include($customView)
     @endisset
 </x-livewire-tables::wrapper>
+@endif
--- a/src/DataTableComponent.php
+++ b/src/DataTableComponent.php
@@ -122,6 +122,16 @@ abstract class DataTableComponent extends Component
     {
         return 'livewire-tables::stubs.custom';
     }
+
+    /**
+     * The view to add any modals for the table, could also be used for any non-visible html
+     *
+     * @return string
+     */
+    public function beforeComponentView(): string
+    {
+        return 'livewire-tables::stubs.before';
+    }
     
     /**
      * @return \Illuminate\Contracts\Foundation\Application|\Illuminate\Contracts\View\Factory|\Illuminate\Contracts\View\View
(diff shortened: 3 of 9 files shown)


lrljoe:
Tested this briefly today, and have a few comments on it.

**beforeComponentView**
There's a new configurable area (**before-tools**), which is positioned just inside the wrapper in _datatable.blade.php_
```php
    <x-livewire-tables::wrapper :component="$this">
        @if ($this->hasConfigurableAreaFor('before-tools'))
            @include(
                $this->getConfigurableAreaFor('before-tools'),
                $this->getParametersForConfigurableArea('before-tools'))
        @endif

```
Is there a reason you'd want another area just above the wrapper?  See [here](https://github.com/rappasoft/laravel-livewire-tables/pull/1037) for the pull request that added this functionality, it was only a week or so ago, so appreciate you've probably not seen it.
Not certain, but it may be better to use the configurable areas if you definitely want to implement another configurable area, just to keep things a little cleaner, although at some point the template file will just be full of configurable areas!

**Hide on Empty**
So a few issues here!
**$rows** contains other things such as the pagination method (if not using simplePagination), so it won't necessarily be empty.  You'd need to adjust datatable.blade.php to have the following instead:
```php
@if (!$this->getHideOnEmptyStatus() || ($this->getHideOnEmptyStatus() && $rows->count() > 0))
```
**However!!!** The big issue with this is if you have no rows as a result of a filter/query, then you're going to end up with the component disappearing at random, and if you've implemented more advanced caching for search/filtering options, then you may end up with a bit of an issue, particularly if you're just an end-user.

So I'd recommend that you adjust the wrapper to cater for if there's an active filter/search in place, maybe something like the following, but please do test more accurately, this will display the table if there's an active filter, a search value, or rows to display.
```php
@if (
    !$this->getHideOnEmptyStatus() ||
        ($this->getHideOnEmptyStatus() &&
            (!empty($this->getSearch()) || 
            $this->hasAppliedFiltersWithValues() || 
            $rows->count() != 0)
         )
   )
```

You should also return something from the blade, otherwise it can be seen as poor practice, and will give a bit of an odd end-user experience!

It is also worth noting that there's already the following option available:
```php
{
  $this->setEmptyMessage('No results found');
}
``` 

Perhaps, taking on board the above feedback around filters/search etc, a "Hide Toolbar on Empty" would be a good addition too as part of this?

gRoberts84:
Hey @lrljoe 

Thanks for looking at my PR and providing feedback.

**beforeComponentView**
The reason for not using the existing before-tools, is that it is still within the main datatable component. 

If I were then to want to hide the component for some reason (e.g. no results) then I'd also have to perform other queries to determine whether to hide what is showing before it (e.g. a title.)

**Hide on Empty**
Apologies, I based this on the @forelse that was being used that had an empty state to show the empty message. 

Admittedly, I'm basing these changes on my use case, where I have a datatable to display all of the entries in a table and we also want to show the 10 upcoming and 10 past entries on a landing page/dashboard.

Initially I had copied the HTML that the component had generated and manually populated it with my data, but it makes sense to reuse the component and disable parts of it to produce the desired effect. 

Although it would be possible, the intention would be to not Hide on Empty whilst allowing the user to filter/search. Maybe we could hide the toolbar if the user enables Hide on Empty?

---

Appreciate you taking the time to look at this for me. 

Happy to discuss in more detail or make changes as and where needed.

lrljoe:
Apologies if I wasn't clear on my point around hide on empty.

With your current code, enable hide on empty on your table, go to an instance that has results, then search or filter in such a way that you get no results. E.g search for loremipsum8888, instead of a "no result" message, the table component will disappear.

This will be an issue if using setSearchDebounce  for example, as if you make a typo when searching and don't correct in time then the component will vanish.  If you have queryString enabled then a refresh of the page won't even return until you clear your queryString.

lrljoe:
> Hey @lrljoe 
> 
> Thanks for getting back to me. 
> 
> I see your point if you were to have both the toolbar and hide on empty enabled. 
> 
> In my use case, I have two instances of a datatable on my dashboard, showing 10 past and upcoming tasks.
> 
> Should the user that is logged in have no tasks to show for either or both, instead of the datatables collapsing and the content below moving up, it will show the No Result.
> 
> I have currently copied the generated HTML and created my own data table at the moment, however I'd rather use the datatable that shows the exact same information etc.
> 
> Thanks again

I actually have the same use case in one of my environments.  
Instead of two tables, I set a public boolean variable to determine whether to show the past or the upcoming tasks.

It was well received by my users so may be worth an a/b test with yours.

My builder just has a when() function to determine which set of tasks to show.

If you'd rather publish the views and customise them as you've done in your fork, then you obviously can do, but just keep an eye out in case of any breaking changes! And use a trait to add your extra functionality.

gRoberts84:
Hey @lrljoe 

Thanks for getting back to me :) 

I'll probably just revert to using my own component with the HTML that was ultimately generated by the datatable.

Apologies if I came across otherwise but I never saw any of your feedback as being negative. 

The whole PR process is meant to work like this and believe me, you've been extremely helpful.

lrljoe:
Just wanted to check whether you've published the views and just made your amends there, then deleted the ones you haven't changed, or if you've done something different?

I do like the idea so I may pilfer your PR and make the necessary adjustments to get it functional when using filters.

PR #1069, opened 2023-02-24
Allow Laravel 10 in V1

DanielGSoftware:
Allows for Laravel 10 in V1.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -22,6 +22,8 @@ jobs:
         exclude:
           - laravel: 10.*
             php: 8.0
+          - laravel: 10.*
+            php: 7.4
           - laravel: 9.*
             php: 7.4
           - laravel: 8.*
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -26,6 +26,9 @@ jobs:
             php: 7.4
           - laravel: 9.*
             php: 7.4
+          - laravel: 9.*
+            php: 8.2
+            stability: prefer-lowest
           - laravel: 8.*
             php: 8.2
             stability: prefer-lowest
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -10,14 +10,18 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         php: [8.1, 8.0, 7.4]
-        laravel: [8.*, 9.*]
+        laravel: [8.*, 9.*, 10.*]
         stability: [prefer-lowest, prefer-stable]
         include:
+          - laravel: 10.*
+            testbench: 8.*
           - laravel: 9.*
             testbench: 7.*
           - laravel: 8.*
             testbench: 6.*
         exclude:
+          - laravel: 10.*
+            php: 8.0
           - laravel: 9.*
             php: 7.4
           - laravel: 8.*
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest, windows-latest]
-        php: [8.1, 8.0, 7.4]
+        php: [8.2, 8.1, 8.0, 7.4]
         laravel: [8.*, 9.*, 10.*]
         stability: [prefer-lowest, prefer-stable]
         include:
@@ -26,6 +26,9 @@ jobs:
             php: 7.4
           - laravel: 9.*
             php: 7.4
+          - laravel: 8.*
+            php: 8.2
+            stability: prefer-lowest
           - laravel: 8.*
             php: 8.1
             stability: prefer-lowest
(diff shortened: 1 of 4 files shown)


lrljoe:
I have much less experience with v1, and I'm not confident in the test suite just yet, so it'll need to be @rappasoft to review it for now.

Apologies, and this will get picked up in due course, your contribution hasn't been ignored, and is definitely appreciated, it's just a bit busy!

nearueng:
Seconding this. It's going to be a pain for us to upgrade all of our current tables, and we'd rather prefer to stay on v1 for now.

Any plans for this soon?

nie7321:
In the v10 contracts package, I don't see anything that would be a breaking change for livewire-tables v1:

https://github.com/illuminate/contracts/compare/v9.52.5...v10.7.1

Some of the type information was moved from the docblock to the method signatures, which is *technically* a BC break. Of those, I do not see any used in the v1 branch:

- `CastsAttributes`
- `CastsInboundAttributes`
- `SerializesCastableAttributes`
- `DataAwareRule`
- `InvokableRule`
- `ValidatorAwareRule`

From framework the upgrade guide, the things that seem potentially impactful:

- [Model "Dates" Property](https://laravel.com/docs/10.x/upgrade#model-dates-property), which is not used in the v1 branch
-  [Query Exception Constructor](https://laravel.com/docs/10.x/upgrade#query-exception-constructor), which is not used in the v1 branch
- [Database Expressions](https://laravel.com/docs/10.x/upgrade#database-expressions) is a little harder to identify (everything else was greppable), but I'm not seeing any case where this would be cast to a string by the v1 branch. Which makes sense -- it's fairly vanilla on what queries it generates, leaving anything complicated up to devs using the package.

nearueng:
I ended up forking it temporarily to make it compatible since it was a blocker.

CyberPunkCodes:
So, does this package currently not work with Laravel 10?

I get no searching, sorting, column selections, or pagination. Like the javascript is dead.

lrljoe:
@CyberPunkCodes 
V2 works fine with Laravel 10.

If you're experiencing an issue like that then it sounds like you're missing Alpine v3, and I suggest you open a discussion.

DanielGSoftware:
Can this be re-opened?

lrljoe:
I've reopened, will try and get to looking at this later this week

lrljoe:
Can I politely request that you also add PHP 8.2 into the workflows, then I should be able to get the tests run and get this merged in.

Apologies for the delay!  I've just not looked at v1 in some time!

DanielGSoftware:
@lrljoe done 😄

lrljoe:
Gah, Apologies.  I've been busy with V3.

I'll re-review this all once we've got a stable V3 out!

lrljoe:
Just noting that we're now onto v3 for this package.  @DanielGSoftware  - always interested in what's stopping an upgrade, do reach out on Discord directly to explain as it may be possible to create a migration tool.

@rappasoft - Think this one can be merged into v1-develop though.  Not had a chance to review breaking changes for v1 for L11 yet, but I know there's at least one or two.

lrljoe:
@DanielGSoftware  - Please can you share an example table from V1?  I don't have a demo to hand, and it'd help me double check that it works before I merge it in.

DanielGSoftware:
> @DanielGSoftware - Please can you share an example table from V1? I don't have a demo to hand, and it'd help me double check that it works before I merge it in.

Sure, what would you exactly like me to share?

DanielGSoftware:
> Just noting that we're now onto v3 for this package. @DanielGSoftware - always interested in what's stopping an upgrade, do reach out on Discord directly to explain as it may be possible to create a migration tool.


Sounds good!. In the past, I opened up a [discussion](https://github.com/rappasoft/laravel-livewire-tables/discussions/888) exactly about why we couldn't upgrade to V2. This has been a while ago, so I'm not sure if these have changed now. I'll probably re-check this later this week. From there, I'll see if contact through Discord is a suitable follow up 😄.

lrljoe:
If you can just share code for a v1 table, with a couple of filters/columns/relations, then that'd be really useful, much appreciated!

Looking at your discussion thread, the key takeaways I found were:

# WithCount()
This does work in v2 & v3

# rowView()
This is not in v2/v3, but you can however specify a view on a per-Column basis.  You can also add attributes to specific rows/columns (based on any value in the row etc)

# Relationships
MorphOne/HasMany are both supported where the package will figure out the relation.
If you use the builder() approach then you can also load in your morphedByMany relations etc, it's just not possible to dynamically load a morphMany type relation without adding the relevant with([]) in the builder method.

Feel free to jump in the Discord, ping me an example of your table and I'll let you know what'll break in v2/v3, and how to get around it.

tdwd:
Requesting this to be reopened!

DanielGSoftware:
> Feel free to jump in the Discord, ping me an example of your table and I'll let you know what'll break in v2/v3, and how to get around it.

Sorry for the inactivity, other things that had priority got in the way. We will probably try to migrate to V3 once it's out of beta, or do some testing and help it get out of Beta as well ;). Probably there shouldn't be any missing features anymore like you mentioned :). If we still encounter something that we feel like it's missing, I will shoot you a message on Discord.

If you still would like me to provide some code from a v1 table, let me know as well and I'll send it over you straight away.

PR #1080, opened 2023-03-03
Filters - FIX - Restrict AppliedFilters to only return non-empty

lrljoe:
At the moment, a filter will be considered "applied" if it has any value, including the default value.  This causes the Pills to appear with a blank value.  To avoid this, I've added an additional check for "isEmpty" in the getAppliedFiltersWithValues.

I am somewhat in two minds on this, but figured that using the filter Defaults for anything other than setting up the array keys (i.e. array with keys, but no values) shouldn't really be done, especially as you can apply filter values at component boot with:
```php
    $this->setFilter('created_after', date('Y-m-d', strtotime('now -1 month')));
```
Tested against the Demo instance.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/Traits/Helpers/FilterHelpers.php
+++ b/src/Traits/Helpers/FilterHelpers.php
@@ -136,7 +136,7 @@ trait FilterHelpers
             ->toArray();
 
         return collect($this->{$this->getTableName()}['filters'] ?? [])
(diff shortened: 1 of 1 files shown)


lrljoe:
This is the most pressing of my current PRs I believe!

lrljoe:
@rappasoft - appreciate you're a busy man, any idea on when you'll be able to take a look at this one please?

PR #1087, opened 2023-03-05
Filters - New Functionality - Adding ID to Filter Labels in menus to allow for easy targeting

lrljoe:
To allow for manipulation/targeting of these elements, either by CSS or JS without impacting on the the wider toolbar elements.  For example, I'd like to add a clickable icon to the label, or an extra attribute to the wrapper.

Also in plan is collapsible filters for when there's a large number in use.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lrljoe:
I'd actually prefer to move the filter labels into the actual blade files for the filters, as that seems a more sensible place to have them than in the toolbar.  Not sure what your thoughts are @rappasoft ?

lrljoe:
Yeah, definitely thinking more it'd be better to have them in with the filter blade, perhaps as an anonymous view component to make maintenance simple.  But I think adding IDs will cause more issues similar to the header/footer filter ID clashing.

lrljoe:
See #1111

Issue #1094, opened 2023-03-09
Header filter inputs doesn't disappear on column hide


taliptako:
Go to demo pages
https://tables.laravel-boilerplate.com/bootstrap-5
Show / Hide the name column to reproduce the error

lrljoe:
Appears across all implementations.

It does disappear after a refresh of the table (e.g. change the sort order).  I'll put a PR together to issue a refresh after changing the Column Selects..

lrljoe:
Interestingly, I'm not getting this issue in my fork, I'll look at where I've fixed it and issue the PR.

lrljoe:
Strangely another of my forks also doesn't have this issue. I suspect that the demo is out of date with current, but will spin up a fresh demo in my env most likely tomorrow to test.

lrljoe:
I've been wholly unable to replicate this using a fresh install of both demo environment & Rappasoft master, I'm assuming that the boilerplate is just out of date.

taliptako:
@lrljoe It seems like it only happens if you activate the filters on the header. Can you also test like that? Using the same configuration as the demo.

lrljoe:
Does the issue happen on this demo (see below) it's a little different to the core demo as it has some custom filters on it, but the codebase is much the same:

[https://tabledemo.lowerrocklabs.com/bootstrap-5](https://tabledemo.lowerrocklabs.com/bootstrap-5)

I've tried with Name filter populated and not populated, showing/hiding the column, and the filter seems to behave.  If it's not happening there, then I'll stick another instance up and see what I did that fixed it.

taliptako:
It's not happening here, but can you also add another column near the name field to make it more similar to the original demo

lrljoe:
Done, and I can at least see the issue now.  Very strange!

lrljoe:
The fix is at least fairly straight forward.  After a quick tweak  to one of the blades, it should then just need a refresh when the Column Select is altered, and that should work.  It's definitely behaving oddly though, and generating multiple of single filters.

lrljoe:
Actually, the fix is even simpler than that, you can't have the same filter in the header & footer, it'll clash, and get itself in a right hissy fit.
If you have have 
Header Filters: Email1, Name1
Footer Filters: Email2, Name2
It works fine

I think a spot of Alpine is needed here to get both fields talking to the same filter.  I'll take a look next week if @rappasoft doesn't beat me to it!

lrljoe:
[...]

PR #1108, opened 2023-03-18
Fix Filter Header/Footer Duplication & Always-Visible Behaviour

lrljoe:
Reference #1094 

When you have the same filter in both header & footer, it will display on screen despite hiding the column, and will shift to another column, and also cause a duplication of that column's filter.

The issue can be seen on the demo instance by 
1) Having two Columns with Secondary Header & Footer Filters (e.g. as in the Name and E-mail Columns)
2) Add anything into one of the Secondary Header/Footer filters
3) Hide that Column.

This appears to be occurring due to there being multiple matching wire:key & ID for the same filter, causing the DOM to become... upset?.

This PR introduces new methods:
setFilterPosition(string $position) // Header or Footer
hasCustomPosition(): bool // Does it Have a filter position set?
getCustomPosition(): string // Retrieves the custom position.

There's a new protected string on the Filter.php to hold the position.

The Custom Position is automatically set by the getSecondaryHeaderContents or getFooterContents functions respectively, the Filter Blades then append either header or footer in an appropriate place in the ID & wire:key respectively, blades have been updated for this.

This allows for the same Filter to exist in multiple places in the page.

There may be a better approach, but this is what I came up with!

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

* [NA] Have you written new tests for your core changes, as applicable?

### DIFF
--- a/resources/views/components/tools/filters/date.blade.php
+++ b/resources/views/components/tools/filters/date.blade.php
@@ -6,8 +6,8 @@
     <div class="rounded-md shadow-sm">
         <input
             wire:model.stop="{{ $component->getTableName() }}.filters.{{ $filter->getKey() }}"
-            wire:key="{{ $component->getTableName() }}-filter-{{ $filter->getKey() }}"
-            id="{{ $component->getTableName() }}-filter-{{ $filter->getKey() }}"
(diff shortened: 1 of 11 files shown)


lrljoe:
See #1111

PR #1114, opened 2023-03-21
Adding basic card layout option

tobiashammarstrand:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/datatable/available-methods.md
+++ b/docs/datatable/available-methods.md
@@ -34,6 +34,19 @@ public function configure(): void
 }
 ```
 
+### setCardLayoutEnabled
+
+Enable card layout instead of table layout.
+
+```php
+public function configure(): void
+{
+  $this->setCardLayoutEnabled();
+}
+```
+
+
+
 ### setTableWrapperAttributes
 
 Set a list of attributes to override on the div that wraps the table
--- a/resources/views/components/card/wrapper.blade.php
+++ b/resources/views/components/card/wrapper.blade.php
@@ -1,0 +1,25 @@
+@aware(['component'])
+@props(['rows', 'columns'])
+
+<ul class="grid grid-cols-3 gap-4 sm:grid-cols-4 md:grid-cols-5 lg:grid-cols-6">
+    
+    @forelse($rows as $row)
+
+    <li wire:key="{{ $loop->index }}" class="col-span-1 bg-white border border-2 border-gray-300 rounded-md shadow cursor-pointer hover:border-af-blue-500">
+
+        @foreach($columns as $colIndex => $column)
+            {{ $column->renderContents($row) }}
+        @endforeach
+        
+    </li>
+
+    @empty
+
+    <li class="col-span-1 bg-white divide-y divide-gray-200 shadow">
+        <div class="flex content-center justify-center p-6">
+            {{ $component->getEmptyMessage() }}
+        </div>
+    </li>
+
+    @endforelse
+</ul>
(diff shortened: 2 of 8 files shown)


tobiashammarstrand:
Will complement with documentation....

dgillier:
Great idea !

lrljoe:
Definitely needs visuals tests adding in!

Also recommend adding in methods to set custom attributes on the card wrapper and card items to keep it consistent with existing.

lrljoe:
Another couple of comments, perhaps add in something like the following:

setLayoutStyle() // options of card or table

Just to keep it consistent with the rest of the code where there's two options, look at the filterLayout for a good example of this.

src/Traits/ComponentUtilities.php
lrljoe, line 1: Should be set to
```php
protected bool $cardLayoutStatus = false;
```

src/Traits/Configuration/ComponentConfiguration.php
lrljoe, line 1: Appreciate they're missing from lots of methods, but please could you add a comment to each of your new methods describing what it does.

I'm currently going through and adding lots of missing/incorrect doctypes/typehints, and it'd be helpful if these were present for new methods!

```php
    /**
     * This sets the layout to use Cards rather than the default Table
     *
     * @param  bool  $status
     *
     * @return $this
     */
    public function setCardLayoutStatus(bool $status): self
    {
        $this->cardLayoutStatus = $status;

        return $this;
    }
```

resources/views/components/card/wrapper.blade.php
lrljoe, line 1: Suggest you add ability to set classes for the
- ul
- li

src/Traits/Helpers/ComponentHelpers.php
```diff
@@ -426,4 +426,12 @@ public function hideConfigurableAreasWhenReorderingIsDisabled(): bool
     {
         return $this->getHideConfigurableAreasWhenReorderingStatus() === false;
     }
+
+    /**
+     * @return bool
```
lrljoe, line 431: As above, would appreciate a comment added here explaining what it does!  It's obvious but it will help in 6 months when someone else looks at it for the first time!

docs/datatable/available-methods.md
```diff
@@ -34,6 +34,19 @@ public function configure(): void
 }
 ```
 
+### setCardLayoutEnabled
```
lrljoe, line 37: To keep it consistent, can you please add a setCardLayoutDisabled() doc as well please

lrljoe:
@tobiashammarstrand  - How's this going, I love the idea by the way, just trying to get everything aligned to the same standards, including the current codebase!

tobiashammarstrand:
I agree on above comments, let's keep it more standardized. I will try get some time during April.

lrljoe:
How's this one looking?  I'm mostly replying so that the PR doesn't autoclose, and I think this is a neat idea!

lrljoe:
@tobiashammarstrand - any news on this one?  I'd hate for your initial hard work to go to waste!

tobiashammarstrand:
I will have more time in some weeks. I agree that we should not waste it.

lrljoe:
Any update on this one?  If it's likely to be stale for some time then that's fine!

lrljoe:
Would this be of use?

dgillier:
At least for me, I think the classic table UI does not really fit mobile use.

So I would love to use the classic layout on Desktop and have a card based layout (ideally with infinite scroll) on Mobile.
Filters and search used on both "modes".

lrljoe:
I'll have a chat with Anthony later and see what he thinks. This is only half done as it stands, but may be a good start.

I do agree, I don't use any tables at all (across the board) on my phone

Issue #688, opened 2022-04-02
Relationship error


jantbox:
There is a problem when you have two relationships with the same table it always shows the first relationship. I'll try to explain:

I have a model like this:

```
class Employee extends Model
{
    ...
    public function company(): BelongsTo
    {
        return $this->belongsTo(Company::class, 'company_id', 'id');
    }

    public function related_company(): BelongsTo
    {
        return $this->belongsTo(Company::class, 'related_company_id', 'id');
    }
    ...
}

```

I have a UserTable component with this piece of code
```
...
Column::make("Id", "id"),
Column::make("Com. Id.", "company_id"),
Column::make("Company", "company.name"),
Column::make("Rel. Id.", "related_company_id"),
Column::make("Rel. Company", "related_company.name"),
...

```
It shows something like this...
<img width="570" alt="livewire_tables_problem_1" src="https://user-images.githubusercontent.com/27139162/161392378-b75d1d59-a2d0-48e8-b88f-5a0232c453f3.png">

The last column in the first line shows "Company 3" when related_company_id = 2. If I remove two first lines, I have something like this

<img width="507" alt="livewire_tables_problem_2" src="https://user-images.githubusercontent.com/27139162/161392425-fc53cf65-8dcb-4327-ad8c-737a190484e3.png">

In this case, the column shows the correct value of the related table.

pablomadariaga:
Rappa I think that to solve this problem, you should allow aliases to be configured for each table that is presented within the query, be it the main table or the relationship tables.

I have a relationship of a column in the same users table to know who invited them to join my platform, and this generates the error:

`SQLSTATE[42000]: Syntax error or access violation: 1066 Not unique table/alias: 'users'`

When trying to perform this query:

`select count(*) as aggregate from `users` left join `users` on `users`.`referrer_id` = `users`.`id``

That is why I think that if the aliases of each table could be configured, the problem would be solved, if there is already a solution at the code level, I would like to receive a bit of your experience.

You did an excellent job, thank you for this wonderful tool.

fabio-ivona:
In V1 there is a merged PR of mine that solves this issue allowing full qualified table names

fabio-ivona:
Yep, I was suggesting to look at here for a new PR addressing the issue

manusiakemos:
```php
Column::make('company', 'company.name')
                ->eagerLoadRelations()

rappasoft:
[...]

PR #1190, opened 2023-05-09
Allow multiple relationships on same table

faeby:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

---

### Introduction
This pull request is an attempt to introduce the possibility to embed two relationships linked to the same table. 

The problem is described in the issue #688. When we try to display two relationships linked to the same table, the first relation will be shown as the second relation too. As there is no alias, the second join is not applied because a join already exists on the target table.

As far as I know, the current workaround consists in configuring the column with the foreign key, and then formatting the column with the `label` or `format` column helper. For me, this has two problems:
- it is not possible to search in the column
- for the second relation, the workaround has to be used, so we have to use two differents methods to display the same thing

### Proposed solution
This PR fix the problem by generating and adding an alias to the tables that are joined. The alias for the relations joins are generated by imploding the relation parts together and glueing them with an underscore. 

Some examples:
I have a document class, with a sender and a producer, where both are a Sender

```php
class Document {
    public function sender(): BelongsTo
    {
        return $this->belongsTo(Sender::class, 'sender_id');
    }

    public function producer(): BelongsTo
    {
        return $this->belongsTo(Sender::class, 'producer_id');
    }
}
```

Table columns configuration:
```php
    public function columns(): array
    {
        return [
            Column::make('sender_id', "sender_id"),
            Column::make('sender.id', "sender.id"),
            Column::make('producer_id', "producer_id"),
            Column::make('producer.id', "producer.id"),
        ];
    }
```

Before the fix:
image

- first line: producer_id is empty, because the document has no producer. But the table incorrectly link the sender 1 as a producer
- second line: producer_id is 2, but the table shows 1, because it is based on the first and unique join

Generated query:
```sql
select 
    `documents`.`sender_id` as `sender_id`,
    `senders`.`id` as `sender.id`,
    `documents`.`producer_id` as `producer_id`,
    `senders`.`id` as `producer.id` 
from 
    `documents` 
left join 
    `senders` on `documents`.`sender_id` = `senders`.`id` -- only 1 join, should have 2 (producer_id is missing)

limit 10 offset 0
```

After the fix
image

Generated query:
```sql
select
    `documents`.`sender_id` as `sender_id`,
    `sender`.`id` as `sender.id`,
    `documents`.`producer_id` as `producer_id`,
    `producer`.`id` as `producer.id` 
from
    `documents` 
left join
    `senders` as `sender` on `documents`.`sender_id` = `sender`.`id` -- alias based on the relationship name
left join
    `senders` as `producer` on `documents`.`producer_id` = `producer`.`id` -- missing join is now present

limit 10 offset 0
```

Because the tables are now aliased, the joins are correctly handled by the `performJoin` method (at line 112).

### Breaking changes
If you provide a custom callback to `Column` `searchable` function and you are specifying a table, you have to replace it by the alias. 

Example: 
In the `tests/Http/Livewire/PetsTable.php` file at line `55`, before the PR is merged:
```php
    ->searchable(
        fn (Builder $query, $searchTerm) => $query->orWhere('breeds.name', $searchTerm)
    ),
```
After the merge:
```php
    ->searchable(
        fn (Builder $query, $searchTerm) => $query->orWhere('breed.name', $searchTerm)
    ),
```
 `breeds.name` was replaced by `breed.name` 

### TODO
- document somewhere the breaking change ?
- add some tests ? Currently there is none for the WithDataTrait

### Discussion
This is my first PR proposal, please let me know if I should change or improve something.

I'd appreciate some feedback about this feature, thank you.

### DIFF
--- a/src/Traits/WithData.php
+++ b/src/Traits/WithData.php
@@ -83,26 +83,34 @@ trait WithData
         }
 
         $table = false;
+        $tableAlias = false;
         $foreign = false;
         $other = false;
+        $lastAlias = false;
         $lastQuery = clone $this->getBuilder();
 
-        foreach ($column->getRelations() as $relationPart) {
+        foreach ($column->getRelations() as $i => $relationPart) {
             $model = $lastQuery->getRelation($relationPart);
+            $tableAlias = $this->getTableAlias($tableAlias, $relationPart);
 
             switch (true) {
                 case $model instanceof MorphOne:
                 case $model instanceof HasOne:
-                    $table = $model->getRelated()->getTable();
-                    $foreign = $model->getQualifiedForeignKeyName();
-                    $other = $model->getQualifiedParentKeyName();
+                    $table = "{$model->getRelated()->getTable()} AS $tableAlias";
+                    $foreign = "$tableAlias.{$model->getForeignKeyName()}";
+                    $other = $i === 0
+                        ? $model->getQualifiedParentKeyName()
+                        : $lastAlias . '.' . $model->getLocalKeyName();
 
                     break;
 
                 case $model instanceof BelongsTo:
-                    $table = $model->getRelated()->getTable();
-                    $foreign = $model->getQualifiedForeignKeyName();
-                    $other = $model->getQualifiedOwnerKeyName();
+                    $table = "{$model->getRelated()->getTable()} AS $tableAlias";
+                    $foreign = $i === 0
+                        ? $model->getQualifiedForeignKeyName()
+                        : $lastAlias . '.' . $model->getForeignKeyName();
+
+                    $other = "$tableAlias.{$model->getOwnerKeyName()}";
 
                     break;
             }
@@ -111,6 +119,7 @@ trait WithData
                 $this->setBuilder($this->performJoin($table, $foreign, $other));
             }
 
+            $lastAlias = $tableAlias;
             $lastQuery = $model->getQuery();
         }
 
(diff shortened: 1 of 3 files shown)


lrljoe:
In terms of  the issue raised in your PR for searching/ordering when using a label/format, I do have a fix ready for a future release (not the pending one), which resolves the problem:

_As far as I know, the current workaround consists in configuring the column with the foreign key, and then formatting the column with the label or format column helper. For me, this has two problems:_
- _it is not possible to search in the column_
- _for the second relation, the workaround has to be used, so we have to use two differents methods to display the same thing_

But for the PR in general, it looks reasonable to me at first glance!
**However** the test cases are going to need to be extended to cater for this type of use case.  

For testing in general, I have been working on extending the test cases, and working on getting them to match the demo repo.  This way, it'll all be a little more straight forward!  I obviously haven't included this type of example yet!

If you are willing to take a look at the Demo repo which is here: [https://github.com/rappasoft/laravel-livewire-tables-demo](https://github.com/rappasoft/laravel-livewire-tables-demo), and are able come up with the model/relationships to demonstrate this, then I'll get it all incorporated into the next iteration of the test cases.  Otherwise there'll be a slightly longer delay while I come up with it (currently I'm working on some AlpineJS functional migrations)

I really, genuinely appreciate any/all contributions, and Anthony definitely does too!

faeby:
Hello @lrljoe,

I've made a draft PR in the demo repo: https://github.com/rappasoft/laravel-livewire-tables-demo/pull/30

I hope it is was what your were asking me to do. Please let me know if it was something else.

---

On another subject, I was thinking about an edge case that should not be supported with this PR. I think we could have a colision with the aliases if we have:
- on the base model, a relation named with an underscore, for example: `my_relation`, the alias will be `my_relation`
- on the base model, a relation named `my`, and this relation has a relation named `relation`, if we try to display a property of the `relation`, its alias will be `my_relation` too

lrljoe:
Thanks, I'll take a look this weekend and look at what needs doing to get it into a future release.

Once I've had a proper look, I'll inevitably hassle you to produce some tests if you haven't already!

lrljoe:
So on reviewing this, my main concern is that it'll be a breaking change for any existing components, as anyone using a relationship column will end up with it failing.

What I'm definitely going to do however is introduce this change into the V3 upgrade which I've just started working on, which is an interim upgrade to make use of the Livewire V3 elements.

lrljoe:
@rappasoft - thoughts on adding this to v3 while we're at it?

rappasoft:
> @rappasoft - thoughts on adding this to v3 while we're at it?

Lets give it a try

lrljoe:
Currently testing in v3.0a

lrljoe:
This is in v3. and is on my backports radar l

PR #1196, opened 2023-05-12
BulkActions - Migration to AlpineJS v2

lrljoe:
### How this works
Changes to a number of blades, and a couple of traits have occurred to allow for this to work smoothly.

The design is so that the "Selected" is only updated on a typical Livewire refresh, rather than forcing it on each checkbox select.

#### Setup
- In wrapper.blade.php, a number of AlpineJS variables are set.
  - selectedCount (_integer_ - used while on the same page to avoid a Livewire refresh)
  - visibleItems (_array_ - stores a list of Visible Items)
  - selectedItems (_entangled_ - stores a list of Selected Items, is a deferred entanglement)
  - allItemsSelected (_boolean_ - returns if all items are selected or not)
  - selectAllOnPage() (_function_  - selects all items on a page)

#### When Checking a Box
- The value of the checkbox is set to the Primary Key of the row, to ensure persistence
- selectedItems (AlpineJS) variable is updated, so it will be updated during the next Livewire lifecycle

#### When loading a new page
- The value of the checkbox is set to the Primary Key of the row, to ensure persistence
- The checkbox for each row is bound to selectedItems, so it will display the correct state when browsing between pages.

### All Submissions:

### New Feature Submissions:

4. [X] Did you test locally to make sure your feature works as intended?

### Changes to Core Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,20 +6,22 @@ All notable changes to `laravel-livewire-tables` will be documented in this file
 
 ### Changed
 - General
-  - Add capability for passing a custom model path to the MakeCommand
-  - Add setFilterDefaultValue() on a per-component basis
-  - Add getFilterDefaultValue() to each Filter Component (to maintain support for PHP7.4 - variable return types)
-  - Moved Setting of Filter Defaults to Traits/Helpers/FilterHelpers - mountFilterHelpers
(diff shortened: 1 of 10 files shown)


lrljoe:
@siebsie23 - This replaces the other PR, and I'd welcome any testing you do before I merge it into develop & then release, as you're clearly keen for this! :)

PR #1202, opened 2023-05-17
added appendRender() to render function

shiv122:
### All Submissions:

### New Feature Submissions:

4. [X] Did you test locally to make sure your feature works as intended?

### Changes to Core Features:

### Introduction

I had a table where i wanted to show sum of all amount on top of table name stats which was loaded using `setConfigurableAreas()`  function . I did not find any way to do that because it should change on every filter and search.

So i added this `appendRender()` to render function so any once can use that to manipulate variables before render get called.


lrljoe:
There are other approaches for extracting values for reuse within the component, so I'd suggest opening a discussion for this instead.

You'd also need to write the relevant tests and documentation for any PR.

lrljoe:
There are setConfigurableAreas([]) to update a specific area on component.

Depending on area, these may be able to access $rows etc.

You can use the Livewire lifecycle hooks where appropriate to update other variables.
[https://laravel-livewire.com/docs/2.x/lifecycle-hooks](https://laravel-livewire.com/docs/2.x/lifecycle-hooks)

Otherwise I'd definitely need to see some documentation, tests etc for this please, and perhaps some use cases if I've misunderstood the intended function?

There's also a conflict with the develop branch which will need resolving.

shiv122:
i know about setConfigurableAreas and i used it . trying doing this=> 

You have a table of subscriptions where you need to get sum of all amount ,
after that you have a Tax calculation class that adds taxes on those ,
i need to access data before to manipulate it and assign it to some public variable so i can use it in   blade which i loaded using `setConfigurableAreas ()`

lrljoe:
I see where you're going with that at least now, thanks for the insight!

Can I ask, what's your plan in terms of what you're going to add to your appendRender() function?  

The reason I ask, is the latest results from the query results that will be displayed to the user won't be available at the point where you're calling your function, as the getRows() is what is actually executing the query.  Anything you call before that point will be one lifecycle behind., and therefore inaccurate I'm afraid!

Where getRows() is called is something I'm looking at separately.

shiv122:
when i use ` $this->sum = $this->getRows()->flatten()->pluck('amount')->sum();` 
what if i want to assign multiple public variables based on this value from getRows(), 
i can not use this in getRows() apart from that if in want to use my service classes to calculate something or  calling external apis based on $value or if i have multiple $variables which depends on getRows() i cant assign those as well. 
Thanks for help man, appreciate it

shiv122:
Thank you so much for help.

PR #1268, opened 2023-07-15
allow hiding database structure using a column alias

KaulSe:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

---

JSON fields or relationships lead to unpleasant query strings and expose unnecessary information to a user.

I propose to add an "alias" like the following examples demonstrates:
```php
Column::make('User Id', 'data->user->id', 'user_id')
    ->sortable(),
```

The URL would look like this when sorting for the user id:
```
?my-table[sorts][user_id]=asc
```

which is much cleaner than the previous version
```
?my-table[sorts][data->user->id]=asc
```

The SQL query will also use the alias;
```sql
select "team_integration_feeds"."data"->'user'->>'id' as "user_id" from "team_integration_feeds" order by "user_id" asc limit 10 offset 0
```

Moreover, this PR should also fix the following error when sorting JSON columns;


lrljoe:
I do like the idea, but I can see a couple of issues, and potentially a better methodology for doing this.

Let me go away and test my theory but I think  this can be done using some of the existing methods without needing much work/modification at all.

KaulSe:
Ok - will try to do it with setCustomSlug() instead :). will push changes asap

lrljoe:
Thanks.

My only other query on it really is - how many people are genuinely pulling back JSON like this, and what are they using for a database backend?

Can you let me know what you're using for example? 
Are you MySQL 5.x, MySQL 8.x, Postgres, Mongo, MSSQL for example?

lrljoe:
You could also use a label here by the way, they're sortable as of the most recent release (about an hour ago), that would also allow you to hide the column(s) from the query string, and you can just use setAdditionalSelects to add the correct column(s) to the query.

KaulSe:
> Thanks.
> 
> My only other query on it really is - how many people are genuinely pulling back JSON like this, and what are they using for a database backend?
> 
> Can you let me know what you're using for example? Are you MySQL 5.x, MySQL 8.x, Postgres, Mongo, MSSQL for example?

I am storing JSON in a JSONB column in a PostgreSQL database in one of my projects. I don't know how many people use this though :).

What do you think about my latest approach? It would still introduce the `alias` variable, but I removed it from the constructor and use the `getAlias()`, `setAlias()` approaches instead.

I had the concern that using the custom slug functionality or the sorting pill title would break existing projects and wouldn't provide this fine-grained control.

lrljoe:
Thanks for the update, I will definitely take a look at it next week.

### This is **definitely** not me saying that I won't merge your PR once I've tested it, as it looks good at first glance.

But - It may be useful for you (for your project directly), to consider using a generated column in Postgres, I'm not as familiar with them as I am with Virtual Columns in MySQL 8, but I'd assume that they behave in a similar way.

For example, you could have a generated column for height in inches that always updates to reflect the correct cm to inch conversion:
```
    height_in numeric GENERATED ALWAYS AS (height_cm / 2.54) STORED
```
MySQL 8.x provides a method for creating a dynamic column from a JSON column, and I'd assume that Postgres has the same.

I would always recommend virtual/dynamic columns from JSON columns as appropriate, as it allows your database engine to index the values, and for you to search using the database directly.

lrljoe:
As an aside, from glancing at your code, it does look like you may be doing some kind of authentication table?  It's worth checking out the [Rappasoft LaravelAuthenticationLog](https://github.com/rappasoft/laravel-authentication-log) package if that's the case, as it provides configurable notifications etc out of the box by monitoring the standard Laravel Events.

lrljoe:
I'm just comparing this and #1190 as they both provide Alias methods.  To avoid any possibility of confusion, I spotted that they've used getTableAlias.

Would you mind swapping your get/setAlias to getColumnAlias and setColumnAlias, as this will almost certainly prevent me from getting mightily confused in the future.
As far as I'm concerned, this PR will take priority over the other, as it has tests/documentation, so it'll be up to the other PR contributor to make additional amends if this PR causes issues.

If you're happy with the above modification, then I can probably get this tested and reviewed sometime next week.

KaulSe:
>It may be useful for you (for your project directly), to consider using a generated column in Postgres, I'm not as familiar with them as I am with Virtual Columns in MySQL 8, but I'd assume that they behave in a similar way.
>As an aside, from glancing at your brief example above, it does look like you may be doing some kind of authentication table?
If so - it's worth checking out the [Rappasoft LaravelAuthenticationLog](https://github.com/rappasoft/laravel-authentication-log) package if that's the case, as it provides configurable notifications etc out of the box by monitoring the standard Laravel Events.

Thank you for the ideas. I have previously used views to "unpack" JSON data in PostgreSQL like this;
```php
Schema::createMaterializedView(
    'team_urls_technologies_view',
    'SELECT jsonb_array_elements_text(technologies) as technology, team_id, asset_id, asset_type, url, id FROM team_urls'
);
```
---

My table itself contains events, so not only authentication logs, but other kinds of data with a different JSON structure. E.g., `user_login`, `user_register`, `user_deleted`, `plugin_installed`, etc., ingested through an API endpoint. Each of these events has its own JSON data structure.

I know I could use some other methods, but in my case, I save a lot of engineering time by just putting everything in a `data` column and directly working with that column using PostgreSQL JSON functions. 
Maybe one day the time will come and I will refactor all of this, but right now, it isn't worth it :).

> Finally, putting my security hat on, security through obscurity is only so valuable (not very!)! If your database is exposed to the internet, fix that first and foremost. Then you can of course consider the Laravel Read DB/Write DB if that's appropriate to reduce the potential for compromise.

I like your security hat, haha. Databases are not accessible from the outside - and I mostly agree with the security through obscurity point.
Though, I still think a bit differently than the major players in the security industry. I believe that making life more difficult for attackers does improve the security posture, but I bet that saying this will get me in trouble, haha.



In that case, I didn't think too much about the security, but more about the accessibility for the user and the looks of the application. Having these nested array elements in the URL `?my-table[sorts][data->user->id]=asc` does look ugly and is not easy to understand for a user. A clean query string improves that.

lrljoe:
Yep, I'm happy with it being a usability rather than security feature.

I'm definitely going to review it later this week and see how it behaves in different demo environments, as it is a great idea.

It should in theory make it into the next release. But may end up in the one after.

KaulSe:
Hey / I will look asap :). ATM in Bulgaria

lrljoe:
Spotted you joined the discord, will pick up with you on there, as it seems to work smoothly in v3 with minimal adjustments at all, so may be a way to get it working in v2 without any headaches

KaulSe:
After talking privately with @lrljoe, he gave me a different solution without merging this PR, which would suit my needs. For other people searching for an alternative method to accomplish it, we can use the `setAdditionalSelects` in the `configure()` method with the SQL keyword `as` to rename a column or table with an alias.

Example:

in the Model:
```php
    protected $casts = [
        'custom_data' => 'json',
    ];
```

in `configure()`:
```php
->setAdditionalSelects(['news.id as id',  'news.custom_data->views as randomDataViews']);
```

In the column configuration:
```php
    Column::make('randomDataViews')
    ->label(function ($row) {
        return $row->randomDataViews ?? 0;
    })
    ->sortable(function (Builder $query, string $direction) {
        return $query->orderBy('custom_data->views', $direction);
    }),
```

The query string looks like (lean & clean query string):
```
?newstablePage=1&newstable-sorts[randomdataviews]=asc&newstable-sorts[name]=asc
```

lrljoe:
Thanks for confirming, I'll look to add this as an example into the docs

PR #1365, opened 2023-09-22
Fix 'clear filters' by using getFilterDefaultValue instead of getDefaultValue

### LINKED ISSUE #1364: [Bug]: Clear filters (setFilterDefaults) clears all set filters, even the default ones
### What happened?

Clicking the 'clear filters' button clears all set filters and does not set it to the default filters. In the code, setFilterDefaults calls resetFilters, which in turn calls setFilters with $filter->getDefaultValue() as a parameter. This always returns null (hard set in the code). I don't think this is correct.

Either the function call should be renamed to setClearFilters (and clear all filters like it does now), or the setFilter called by resetFilters should call $this->getFilterDefaultValue().

### How to reproduce the bug

Click clear filters button when filters are applied.

### Package Version

2.15

### PHP Version

8.1.x

### Laravel Version

10.24

### Alpine Version

3.13

### Theme

Bootstrap 5.x

### Notes

_No response_

### Error Message

_No response_

RenoLooijmans:
Fix 'Clear filters' to reset the filters to their default values instead.

Please be aware this breaks expected functionality for other users, as filters are reset to their default values instead of cleared all together. In past versions the filters were also reset to their default values.

Fixes #1364

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/src/Traits/Helpers/FilterHelpers.php
+++ b/src/Traits/Helpers/FilterHelpers.php
@@ -222,7 +222,7 @@ trait FilterHelpers
             $filter = $this->getFilterByKey($filter);
         }
 
-        $this->setFilter($filter->getKey(), $filter->getDefaultValue());
+        $this->setFilter($filter->getKey(), $filter->getFilterDefaultValue());
     }
 
     public function getFilterLayout(): string


lrljoe:
Hi @RenoLooijmans 

Thank you for your commit/PR!  Is this PR targeted at V2 or V3?

Either way, please can I suggest that you do the following:

- Add a boolean to an appropriate place (either Per-Filter, or Per-Component)
- Add a get/set method in the appropriate Helper & Configuration traits to update the boolean (again, per-filter, or per-component)
- Use the get() method defined in the helper to determine whether to use the value from getFilterDefaultValue or getDefaultValue

This way, existing implementations won't change, but it adds a valuable option for anyone who wants it.

I'm happy with either a "global" or "per-filter" approach, perhaps per-filter would offer more configurability?

Issue #1367, opened 2023-09-23
[Bug]: assertViewIs is not working


jacksontong:
### What happened?

`assertViewIs` always return `The response is not a view.` for normal Laravel route test, however it's working fine for Livewire component. If I set `inject_assets` to false then it's working again.

### How to reproduce the bug

```php
class DocumentController
{
    public function index(): View
    {
        return view('documents.index');
    }
}

it('renders correct view', function () {
    \Pest\Laravel\get(route('documents.index'))
        ->assertOk()
        ->assertViewIs('documents.index');
});
```

### Package Version

v3-develop

### PHP Version

8.2.x

### Theme

None

lrljoe:
This should be resolved in the latest v3-develop.  I adjusted your PR somewhat, and also threw a load of new tests and amendments to behaviour in at the same time.

Please do let me know your findings!

jacksontong:
thanks, it's fixed on my end.

PR #1368, opened 2023-09-23
Fix assertViewIs is not working

jacksontong:
### All Submissions:

### Changes to Core Features:

This is a fix for #1367 

https://github.com/rappasoft/laravel-livewire-tables/blob/v3-develop/src/Features/AutoInjectRappasoftAssets.php#L60
`$handled->response->setContent(static::injectAssets($html));` overrides `$original` property of response which `assertViewIs` uses to assert the response.

https://github.com/laravel/framework/blob/10.x/src/Illuminate/Testing/TestResponse.php#L1131C30-L1131C30
```php
protected function responseHasView()
{
    return isset($this->original) && $this->original instanceof View;
}

### DIFF
--- a/src/Features/AutoInjectRappasoftAssets.php
+++ b/src/Features/AutoInjectRappasoftAssets.php
@@ -57,7 +57,9 @@ class AutoInjectRappasoftAssets extends ComponentHook
             $html = $handled->response->getContent();
 
             if (str($html)->contains('</html>')) {
+                $original = $handled->response->original;
                 $handled->response->setContent(static::injectAssets($html));
+                $handled->response->original = $original;
             }
         });
     }

### CODE REVIEW

**Comment by lrljoe** [comment]
Will enable the tests, give it a quick run-through in my test environment, then merge in if all working.

Thank you for the PR!!

From speaking to the core team, we're the only large package using this functionality at the moment, and I am finishing off some additional toggles etc to make disabling it easier for those who'd rather use the app.js approach to include the relevant files.

**Comment by lrljoe** [comment]
May need to do a quick edit to your PR so that the methods are checked for prior to use, and that should prevent any oddity with it running and methods/props not being available (which is what PHPStan has complained about)

Will check in my dev env, but this should be merged shortly

**Comment by lrljoe** [comment]
I've made a couple of minor changes to your change in a separate branch, and doing some final checks.

Note that core Livewire also uses the same methodology, will check if they've tweaked their implementation since I last reviewed it.

Main changes that I've made in a separate branch:
1. Replaced
```
                $original = $handled->response->original;
```
with
```
                $original = $handled->response->getOriginalContent();
```
As this uses the standard methods, rather than direct properties (matching the typical style of the project).

2. Added checks for: 
presence of method "getOriginalContent"
presence of property "original"
To ensure that non-existent methods/properties aren't called

3. TypeHinted the $handled parameter (needed doing anyway)

Doing some more checks, but it now passes PHPStan, and existing Test Suite.

lrljoe:
Fix merged in separate PR, with amends

PR #1390, opened 2023-10-05
Update DataTableComponent.php

Traqza:
### All Submissions:

* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/DataTableComponent.php
+++ b/src/DataTableComponent.php
@@ -73,6 +73,7 @@ abstract class DataTableComponent extends Component
 
     /**
      * Runs on every request, after the component is mounted or hydrated, but before any update methods are called
+     * @throws DataTableConfigurationException
      */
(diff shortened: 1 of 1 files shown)


lrljoe:
So we made a conscious decision a while back to strip out inline documentation from the core code.

Previously, every method had an explanation, typehint explanation etc.

As v3 is PHP8+, we moved to proper typehinting, for attributes and methods.

There are a handful of docs inline remaining, and one or two pieces remaining for cleanup.

Happy to discuss further, but it'd feel odd to start adding in pieces that don't add value to the actual application.

There is also a vague plan for a full API reference at some point, but that's some way off

PR #1430, opened 2023-10-19
v3 - Full Page Table Component Options

### LINKED ISSUE #1453: [Feature Request]: Support for using a DataTableComponent as a Full Page Component
### Overview

Add a support for rendering full page component instead of needing to wrap the Table Component it in a view.

### Detailed explanation

Can directly reference a component that extends DataTableComponent via a route, without needing to use an interim view/component.

### Notes

_No response_

amshehzad:
This PR adds different configuration options for setting up a full-page component for livewire
https://livewire.laravel.com/docs/components#full-page-components

The following methods have been added:

- `setLayout` for setting layout for component (https://livewire.laravel.com/docs/components#per-component-layout-configuration)
- `setSlot`
- `setExtends`
- `setSection`

### DIFF
--- a/src/DataTableComponent.php
+++ b/src/DataTableComponent.php
@@ -129,11 +129,30 @@ abstract class DataTableComponent extends Component
         $this->setupFooter();
         $this->setupReordering();
 
-        return view('livewire-tables::datatable')
-            ->with([
-                'columns' => $this->getColumns(),
-                'rows' => $this->getRows(),
-                'customView' => $this->customView(),
-            ]);
+        $view = view('livewire-tables::datatable');
+
+        if (isset($this->layout)){
+            $view->layout($this->layout);
+        }
+
+        if (isset($this->extends)){
+            $view->extends($this->extends);
+        }
+
+        if (isset($this->section)){
+            $view->section($this->section);
+        }
+
+        if (isset($this->slot)){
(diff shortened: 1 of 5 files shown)


lrljoe:
Only other thing that may be really useful to add for enabling full page components is:
```
- setHeaders(array $headers)
```

As Livewire 3 offers this now:
```
    public function render()
    {
        return view('livewire.show-post')
            ->response(function(Response $response) {
                $response->header('X-Custom-Header', true);
            });
    }
```

So a full page component should offer the ability to add in headers.

Ultimately, plan is to expose render(), but I'm 99% that's a v4 thing.

amshehzad:
@lrljoe I've added the helper methods as you suggested, but havn't added the title and headers method, will do them in a new pr

lrljoe:
Sounds good, I'll review this weekend as have a few PRs to review and merge.

lrljoe:
Few conflicts, but that's due to other changes that have happened since you first forked, and are very recent (such as adding hook methods, moving bits out of rendering etc).

At this point, I imagine I'll end up moving your code into a trait of its own for a "WithFullPageComponentMethods" or similar.

Sadly this PR has come in right when I'm moving bits around, but it definitely won't go to waste!

lrljoe:
Replaced by #1580

PR #1642, opened 2024-01-14
Add support to RTL language  

hussenk:
There is a problem with the header of the table it is always on the left. so I added config 'dir' to show where it's supposed to be.

* [ ] Have you successfully run tests with your changes locally?
- yes

It's working in Tailwind CSS, I don't use Bootstrap.


Pezhvak:
+1

lrljoe:
Will have to add in bootstrap, so can't merge until that's done, should have time to look at that next weekend
Would also recommend that this is done locally to the table to allow for on the fly swapping of localisation, rather than hard configured.

You can presently do this with the setThAttributes() however!

PR #1690, opened 2024-03-16
UI patch: reorder if statement in toolbar configurable area to fix UI inconsistencies

itsLeonB:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

---

#### Before patch
before-patch
Here we can see the search bar have a `margin-left` on it, it is because the parent div with style `md:space-x-2` is making a margin in between child elements. The `toolbar-left-start` area is rendered with 0 width, therefore the margin.
#### After patch
after-patch
Here the search bar is properly aligned.

#### Before patch
before-patch
Here the button I added for `toolbar-left-start` has a different height than other toolbar elements. This is caused by the parent div does not have the proper classes unlike the other toolbar elements.
after-patch
After adding the classes like other toolbar elements, the button now has the proper height.


lrljoe:
Apologies, been a bit busy of late!  Will look to get this reviewed and merged in this week.

PR #1714, opened 2024-05-03
Scan provided path for all available FQCNs in paths within and outside app folder

marvoh:
As we work on newer projects, we find that the scope of functionality begins to become repetitive and therefore you or someone else may have already done some piece of work that you already need. These would typically be installed in the vendors folder through composer.

Some of these packages have models which the user of the end product would need to interact with therefore will require presentation in tables. However, this package at the moment does not accept packages outside app path.

If time permits and resources permits, please consider looking into supporting both scenarios.


lrljoe:
So will definitely review this in the next week.

Typically I've seen people extending a Vendor Model, rather than directly referencing it, but happy to adjust the Make Command to allow for this
l

marvoh:
The breakages are strange also to me. I can stick to best practise and extend the models instead?

lrljoe:
Can you resolve the conflict in CHANGELOG.md, and let's see if the tests re-run and pass, as they seem to be mostly passing locally (I had one error when running it in a test environment)

lrljoe:
This will be in the next release (should be this weekend).

PR #1742, opened 2024-06-20
fix: Apply cursor pointer only on clickable columns when using Bootst…

MP70:
### All Submissions:

- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### Issue Fix:

- [x] **What issue does this address?**
  - This change resolves an issue with the cursor styling on table rows **when using Bootstrap**. Previously, `cursor: pointer;` was always applied to all cells, regardless of whether the column was set as clickable. This contradicts the documentation and expected behavior. With this update, the cursor changes to a pointer only when the column is explicitly marked as clickable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `laravel-livewire-tables` will be documented in this file
 
+## [v3.2.7] - 2024-06-05
+### Bug Fixes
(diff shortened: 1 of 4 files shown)


lrljoe:
Will give this a once over and merge, either this weekend, or early next week.

Apologies for the delay!

PR #1807, opened 2024-07-31
Fix set filter event for multi selectables

edwinvdpol:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/Traits/WithEvents.php
+++ b/src/Traits/WithEvents.php
@@ -14,7 +14,7 @@ trait WithEvents
         $this->clearSorts();
     }
 
-    public function setFilterEvent(string $filter, string $value): void
+    public function setFilterEvent(string $filter, mixed $value): void
     {
         $this->setFilter($filter, $value);
     }


lrljoe:
It should also work with this which avoids the "mixed", if it works, then update the commit
```
    public function setFilterEvent(string $filter, string|array $value): void
```

edwinvdpol:
That should also work, but I'd rather keep this in sync with `setFilter()` in `FilterHelpers.php` ([here](https://github.com/rappasoft/laravel-livewire-tables/blob/development/src/Traits/Helpers/FilterHelpers.php#L135)). I will change it anyway.

lrljoe:
Feel free to update the other one as well, happy to have both matching

lrljoe:
I'll merge this in later today, and will look at removing the mixed at a later date if not updated by the time I look at it.

Just did two releases in short-order, so it probably won't actually get released until the middle of the week

lrljoe:
This is now in the latest release, it now accepts, string|array|null to allow for the three defined scenarios

PR #1900, opened 2024-08-26
Cleanup helpers

edwinvdpol:
### All Submissions:

* [ ] Have you followed the guidelines in our Contributing document?

### New Feature Submissions:

### Changes to Core Features:


lrljoe:
Hi @edwinvdpol - just before I start reviewing this, can you give some context please!

lrljoe:
Haven't looked at what test failed, but please ensure they pass locally as well!

lrljoe:
Also - rather than removing Doc Blocks - please update them to a more accurate definition.  The plan is to add fuller descriptions for all methods, rather than removing the more generic ones.

lrljoe:
> Not sure why, but the `development` branch is 2 commits behind on `master` branch?
> 
> I am not sure why there is a conflict, when I merge `development` in my branch, it is up-to-date. Is everything merged into `development` properly?
> 
> There is also a `develop` branch? Not sure what the exact workflow should be to submit a PR...

"development" is the correct development branch.  "develop" is just a hangover as it's got more protection rules on it that mean I can't easily update the changelog etc

The reason for the 2 commits behind is just the styling/cleanup when it merges into master from development, and changelog  gets updated, there's nothing missing.

lrljoe:
The Merge Conflict is due to an update from earlier today which updated the ActionsHelpers, so you'll just need to resolve the conflict.

lrljoe:
At a glance, I can see a couple of issues, the changes to BulkActionsHelpers means that some of the logic won't work as expected.

For example: this won't be respected:
```
hideBulkActionsWhenEmptyIsEnabled()
```

lrljoe:
A lot of this is fine at a glance, but spotting that one has made me want to do a more thorough review.

src/Traits/Helpers/BulkActionsHelpers.php
```diff
@@ -62,47 +62,33 @@ public function hasBulkActions(): bool
...
 
-    /**
-     * @return array<mixed>
-     */
     public function getBulkActions(): array
     {
         return $this->bulkActions();
     }
 
     public function showBulkActionsDropdown(): bool
```
lrljoe, line 70: Needs Review

lrljoe (changes requested):
Please see comments, most is fine tho and appreciated.

edwinvdpol:
Weird that GH actions in failing, but not locally...

Screenshot 2024-08-27 at 11 07 54

edwinvdpol:
> At a glance, I can see a couple of issues, the changes to BulkActionsHelpers means that some of the logic won't work as expected.
> 
> For example: this won't be respected:
> 
> ```
> hideBulkActionsWhenEmptyIsEnabled()
> ```

Can you elaborate what you mean?

lrljoe:
> > At a glance, I can see a couple of issues, the changes to BulkActionsHelpers means that some of the logic won't work as expected.
> > For example: this won't be respected:
> > ```
> > hideBulkActionsWhenEmptyIsEnabled()
> > ```
> 
> Can you elaborate what you mean?

Your code breaks the "hideBulkActionsWhenEmptyIsEnabled" behaviour:

```
    public function showBulkActionsDropdown(): bool
    {
        if (! $this->bulkActionsAreEnabled()) {
            return false;
        }

        if ($this->hasBulkActions()) {
            return true;
        }

        if ($this->hideBulkActionsWhenEmptyIsEnabled()) {
            return $this->hasSelected();
        }

        return false;
    }
```

So if it "hasBulkActions"
```
        return count($this->bulkActions()) > 0;
```
Then it will return true, regardless of hideBulkActionsWhenEmptyIsEnabled status.

So any Table that "hasBulkActions would return true.

hideBulkActionsWhenEmptyIsEnabled prevents the "bulk actions" dropdown from being displayed, IF there are no rows selected.  Your code would break this, as it would never be called.

PR #1905, opened 2024-08-27
Updated test workflows

edwinvdpol:
This will remove duplicate tests entry for different Laravel versions and removes parallel testing to prevent weird errors.


lrljoe:
Will need to consider the impact of this.

The decision for segregating the Laravel versions was recent, and was purely to make it easier to run test sections.

Removing the multi-threading, while it **may** clear up the sporadic errors when there are lots of PRs running simultaneously, will significantly slow down the overall test speed.

I'd rather have to re-run a failed test occasionally than wait 15 minutes for tests to complete.

PR #1993, opened 2024-10-03
Fix missing i18n keys  for FR, IT, EN, DE

khwadj:
This PR aims at fixing the missing localization keys throughout the language files.

For the few languages I more or less speak (french, spanish, italian, german), those new keys have also been translated.

### All Submissions:


lrljoe:
Will take a look this weekend to get this merged in. Don't worry about the test that failed, it'll pass when I rerun it

lrljoe:
Okay, having reviewed this PR - please don't add in the English localisations into the non-English locale files.

The reason being - I do contact those who contribute localisations/translations when I spot an unlocalised string, and I won't spot it with this!

It should fail back to English if there's no translated key present.

Please update your PR, so that you're only adding missing keys that are valid localisations.

khwadj:
The fallback language is defined at the app level. In our case, that's french, so missing keys will have an impact (see image), and it will for anybody not defining EN as their fallback language at the app level, since I've found missing keys in all language files.

I'm okay fixing only a few languages if that's what you want, since it's ultimately what serves me. 


_Note: I'm having an issue with fallback locale not working, but it might be the laravel version we're using (v10.48.12, I plan on ugrading later this month to the latest one), and I don't have much time available at the moment to investigate it, so let's ignore that._



On a related but different subject, some keys would benefits from being reworked as single strings with variables rather that several ones. "Showing X of Y on Z results" is an english sentence and translates very poorly in other languages when maintaining this syntax. Defining a single translation key with X, Y, and Z variables would gracefully solve this issue by allowing each language to phrase it properly. Let me know if you're open to a discussion and/or change relatively to this aspect.

lrljoe:
> On a related but different subject, some keys would benefits from being reworked as single strings with variables rather that several ones. "Showing X of Y on Z results" is an english sentence and translates very poorly in other languages when maintaining this syntax. Defining a single translation key with X, Y, and Z variables would gracefully solve this issue by allowing each language to phrase it properly. Let me know if you're open to a discussion and/or change relatively to this aspect.

Very happy to have a discussion around that.  I believe from memory that the translation key for "Showing X of Y on Z results" mirrors the default Livewire/Laravel behaviour, but more than happy to amend.

I am trying to keep track of the contributors for localisations on the Discord, so do poke me if you haven't already.

PR #2201, opened 2025-02-15
Bulk actions attribute tweaks

lrljoe:
Replaced by #2203


lrljoe:
Still in progress, there's some standardisation and tweaking being added, as well as a core checkbox blade to avoid repeated code/classes.

lrljoe:
Replaced by #2203

PR #2209, opened 2025-02-20
Added external DIV classes to avoid misalignement with column-select …

inforisorse:
Added style classes in order to correctly align to column select and search box.


rappasoft:
Thanks @inforisorse — and sorry this sat for so long. The stale bot closed it before it was ever properly reviewed, which was our failure, not yours.

Your contribution has now shipped in [v3.8.0](https://github.com/rappasoft/laravel-livewire-tables/releases/tag/v3.8.0): **The Tailwind per-page dropdown is now aligned with the column-select beside it.**

Closing as landed.

PR #2231, opened 2025-03-05
Refactor pagination logic to ensure consistent total item count handling

mattusik:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

---
# ✨ Enhancement: Cursor Pagination Without Count

## Summary  
This PR extends the package by enabling cursor pagination to function without requiring a total count. While this feature was already available for simple pagination, it was not implemented for cursor-based pagination.

## Problem  
Simple pagination and cursor pagination differ in how they retrieve records:  

### **Simple Pagination** (Uses `LIMIT` and `OFFSET`)  
```sql
SELECT * FROM table LIMIT x OFFSET y;
```
- This approach requires counting total records, which can be slow for large datasets.
- Using OFFSET can cause performance issues as the database must scan all previous rows.

### **Cursor Pagination** (Uses id > {cursor} and LIMIT)
```sql
SELECT * FROM table WHERE id > {cursor} LIMIT x;
```
- More efficient because it avoids scanning previous rows.
- Does not require a total count, making it faster on large datasets.

## Solution ##

- Added support for cursor pagination without requiring a total count.
- This allows users to benefit from cursor pagination’s efficiency without unnecessary database load.

## Impact ##

- Improves performance, especially for large datasets.
- Maintains backward compatibility with existing implementations.

### DIFF
--- a/src/Traits/WithData.php
+++ b/src/Traits/WithData.php
@@ -107,17 +107,23 @@ trait WithData
                     $this->paginationTotalItemCount = $this->getBuilder()->count();
 
                     return $this->getBuilder()->simplePaginate($this->getPerPage() === -1 ? $this->paginationTotalItemCount : $this->getPerPage(), ['*'], $this->getComputedPageName());
-                } else {
-                    $this->paginationTotalItemCount = -1;
-
-                    return $this->getBuilder()->simplePaginate($this->getPerPage() === -1 ? 10 : $this->getPerPage(), ['*'], $this->getComputedPageName());
                 }
+
+                $this->paginationTotalItemCount = -1;
+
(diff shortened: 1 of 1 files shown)


rappasoft:
Thanks @mattusik — and sorry this sat for so long. The stale bot closed it before it was ever properly reviewed, which was our failure, not yours.

Your contribution has now shipped in [v3.8.0](https://github.com/rappasoft/laravel-livewire-tables/releases/tag/v3.8.0): **Cursor pagination now honours `setShouldRetrieveTotalItemCountDisabled()` — avoiding the `count()` is the whole point of cursor pagination, and the toggle was only ever checked for simple pagination.**

Closing as landed.

PR #2278, opened 2025-07-02
Only aggregate selected columns

glamax13:
### All Submissions:

### New Feature Submissions:

1. [-] Does your submission pass tests and did you add any new tests needed for your feature?
2. [-] Did you update all templates (if applicable)?
3. [-] Did you add the [relevant documentation](https://github.com/rappasoft/laravel-livewire-tables-docs) (if applicable)?

### Description
Aggregation can complicate the query quite a bit and significantly slows down the query.
I would have like it if only "usefull" aggregate columns where calculated.

I understand that filters and sort could be applied even on invisible columns so this would break current behaviour.
A global setting to enable and disable this behaviour could be preferable and i don't know how "hiddenColumn" should interract with this.
In my case the potential performance gains justify this feature usefullness but i let you be the juge on this matter.

This change is small enough that i didn't feel the need to add test or documentation changes.
Let me know if they are required and if other matters should be taken into concideration.

Thank you for your time


rappasoft:
Thanks @glamax13 — and sorry this sat for so long. The stale bot closed it before it was ever properly reviewed, which was our failure, not yours.

Your contribution has now shipped in [v3.8.0](https://github.com/rappasoft/laravel-livewire-tables/releases/tag/v3.8.0): **Aggregate subqueries are skipped for deselected columns when `setExcludeDeselectedColumnsFromQuery()` is on.**

Closing as landed.

PR #2286, opened 2025-07-29
Add example to filters-boolean.md

Muetze42:
I was slightly surprised that the “classic” example is not there.

### All Submissions:

No Changes, just a little example for the docs.


rappasoft:
Thanks @Muetze42 — and sorry this sat for so long. The stale bot closed it before it was ever properly reviewed, which was our failure, not yours.

Your contribution has now shipped in [v3.8.0](https://github.com/rappasoft/laravel-livewire-tables/releases/tag/v3.8.0): **The plain BooleanFilter form is now documented.**

Closing as landed.

PR #2318, opened 2026-02-17
Upgrade to Livewire 4

DanielGSoftware:
# Upgrade Summary

Made with the help of Claude Sonnet 4.6. I fed it the upgrade guide, made a plan, review the plan and made sure to adjust some things I thought was necessary.

## Changes
- Bump `livewire/livewire` constraint from `^3.0|dev-main` → `^4.0`
- Refactor `MakeCommand` to remove deleted Livewire v3 internals
- Update `wire:model.blur` → `wire:model.live.blur` to preserve behavior
- Update documentation to reflect the `.live.blur` change

---

### MakeCommand (`src/Commands/MakeCommand.php`)

Livewire v4 removed:

- `ComponentParser`
- `LivewireMakeCommand`

Both were previously part of the `SupportConsoleCommands` feature.

The `make:datatable` command depended on them for:

| Dependency | Purpose |
|----------|------|
| `ComponentParser` | Resolve class name, namespace, and path from user input |
| `LivewireMakeCommand::isReservedClassName()` | Validate reserved component names |

### Replacement Implementation
A self-contained implementation now:

- Resolves class name using `Str::studly()`
- Resolves namespace and file path using:
  - `livewire.class_namespace`
  - `livewire.class_path`

The reserved class name validation was **removed**.

> It only prevented names like `Component` or `Template`, which are unrealistic for datatable generation.

---

### Livewire v3 behavior
- `wire:model` → deferred by default
- `wire:model.blur` → syncs value when the field loses focus

### Livewire v4.1 behavior
- `wire:model` → still deferred
- `wire:model.blur` → **no longer works**
- `.blur` now only applies to **live models**

I thought it would make most sense in this PR to stick to the old behaviour, so all `blur` references were updated to `live.blur`. This should still be compatible with v4.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "blade-ui-kit/blade-heroicons": "^2.1",
         "illuminate/contracts": "^10.0|^11.0|^12.0",
         "illuminate/support": "^10.0|^11.0|^12.0",
-        "livewire/livewire": "^3.0|dev-main"
+        "livewire/livewire": "^4.0"
     },
     "require-dev": {
         "ext-sqlite3": "*",
--- a/docs/filters/available-filter-methods.md
+++ b/docs/filters/available-filter-methods.md
@@ -92,7 +92,7 @@ For the following Filters, you may customise how the input is wire:model into th
 You may override this using the following methods, on any of the above Filter types:
 
 ### setWireBlur()
-Forces the filter to use a wire:model.blur approach
+Forces the filter to use a wire:model.live.blur approach
 ```php
     TextFilter::make('Name')
     ->config([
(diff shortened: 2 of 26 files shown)


danie-ramdhani:
image

ChandanShakya:
https://github.com/rappasoft/laravel-livewire-tables/pull/2318#issuecomment-3919046489

@danie-ramdhani Could you ask for eta?

ChandanShakya:
With livewire v4 update could you include [blaze](https://github.com/livewire/blaze) built in for this project

DanielGSoftware:
> With livewire v4 update could you include [blaze](https://github.com/livewire/blaze) built in for this project

That'd be awesome, I can imagine there will be quite some performance gains. However, that should be done in a separate PR, it's not tied to Livewire 4. Blaze is a standalone package. 

Livewire 4 does however also bring more tools for optimising performance. I don't know this package too well, but I have the feeling there might be some cases that are eligible for this. However, I think ideally this optimisation should be done in a follow up PR and should not be blocking this PR, so that firstly we can just upgrade our projects to Livewire 4.

